PR #65, opened 2015-04-21
Added "brew tap" command for OS X install.

kubu4:


benvvalk:
Thanks!

sjackman:
Thanks, for the patch, @kubu4. Even tidier is `brew install homebrew/science/abyss`

PR #69, opened 2015-05-17
some fixes for 32bit builds

mcirsta:
This fixes part of the problems for 32bit builds ( issue 68 )

### DIFF
--- a/Bloom/Bloom.h
+++ b/Bloom/Bloom.h
@@ -23,6 +23,7 @@
 #include "Common/IOUtil.h"
 #include "DataLayer/FastaReader.h"
 #include <iostream>
+#include <stdint.h>
 
 #if _OPENMP
 # include <omp.h>
@@ -51,7 +52,7 @@ namespace Bloom {
 	/** file format version number */
 	static const unsigned BLOOM_VERSION = 2;
 	/** I/O buffer size when reading/writing bloom filter files */
-	static const unsigned long IO_BUFFER_SIZE = 32*1024;
+	static const uint64_t IO_BUFFER_SIZE = 32*1024;
 
 	/**
 	 * How to treat existing bits in the bloom filter when
@@ -165,7 +166,7 @@ namespace Bloom {
 		// bloom filter bits
 
 		size_t bits = endBitPos - startBitPos + 1;
-		size_t bytes = (bits + 7) / 8;
+		uint64_t bytes = (bits + 7) / 8;
 		char buf[IO_BUFFER_SIZE];
 		for (size_t i = 0, j = 0; i < bytes;) {
 			size_t writeSize = std::min(IO_BUFFER_SIZE, bytes - i);
--- a/Common/StringUtil.h
+++ b/Common/StringUtil.h
@@ -6,6 +6,7 @@
 #include <sstream>
 #include <string>
 #include <cmath>
+#include <stdint.h>
 
 /** Return the last character of s and remove it. */
 static inline char chop(std::string& s)
@@ -106,7 +107,7 @@ bool isReadNamePair(const std::string& name1, const std::string& name2)
 	return false;
 }
 
-static inline size_t SIToBytes(std::istringstream& iss)
+static inline uint64_t SIToBytes(std::istringstream& iss)
 {
 	double size;
 	std::string units;
@@ -133,19 +134,19 @@ static inline size_t SIToBytes(std::istringstream& iss)
 
 	switch(tolower(units[0])) {
 		case 'k':
-			size *= (size_t)1<<10; break;
+			size *= (uint64_t)1<<10; break;
 		case 'm':
-			size *= (size_t)1<<20; break;
+			size *= (uint64_t)1<<20; break;
 		case 'g':
-			size *= (size_t)1<<30; break;
+			size *= (uint64_t)1<<30; break;
 		case 't':
-			size *= (size_t)1<<40; break;
+			size *= (uint64_t)1<<40; break;
 		default:
 			iss.setstate(std::ios::failbit);
 			return 0;
 	}
 
-	return (size_t)ceil(size);
+	return (uint64_t)ceil(size);
 }
 
 static inline size_t SIToBytes(const std::string& str)


sjackman:
My preference would be to consistently use `size_t` everywhere, and hack around that `1<<40` issue.

sjackman:
Replacing `1<<40` with `1ULL << 40` might do the trick.

benvvalk:
Yes, I like using size_t too.  Alternatively, I could just take out the option of using a "T" suffix for terabyte-size Bloom filter sizes.  That might fix it too.

mcirsta:
The problem with size_t is that there's no guarantee of the size of size_t whereas uint64_t guarantees a 8 byte size no matter what.
 What I do consider bad practive are operations like in Bloom/Bloom.h where an unsigned long is compared to size_t. That only happened to work on 64bit but it will probably work nowhere else and I do find uint64_t provides consistency here.

 Is there a reason to avoid uint64_t ? It's definde in  C++11 as far as I know.

sjackman:
`size_t` is sufficiently wide to address the maximum amount of memory on the host machine. On a 32-bit machine `size_t` will be more efficient than `uint64_t`. A `uint64_t` is unnecessarily wide, since the host machine is only able to address a maximum of 4 GB of RAM.

mcirsta:
I fully agree with this but it's all down to what values you inted to keep in that variable. For memory locations of size of a string and other things like that size_t is more than enough but if for instance you want to keep track of the number of sand particles on a beach there's not guarantee a 32 bit size_t will be large enough ( well there's not guarantee a 64 bit one will be but this is just an example ).
 
 I've no idea what values you have in this software and how large they can be ... if they're not larger than max(4 byte) and will not be so in the future then indeed size_t is good enough.

sjackman:
We do in fact use `uint64_t` when we're counting, for example, the number of *k*-mer observed, but use `size_t` when counting the number of distinct *k*-mer, which can't exceed the amount of addressable memory.

sjackman:
In any case, your PR certainly helps highlight areas of code that need to reviewed to see if they've used the right type (`uint64_t` or `size_t`). We may not merge this PR, but we'll commit something to fix your compiler error.

Bloom/Bloom.h
```diff
@@ -51,7 +52,7 @@ namespace Bloom {
 	/** file format version number */
 	static const unsigned BLOOM_VERSION = 2;
 	/** I/O buffer size when reading/writing bloom filter files */
-	static const unsigned long IO_BUFFER_SIZE = 32*1024;
+	static const uint64_t IO_BUFFER_SIZE = 32*1024;
```
sjackman: Use `size_t`.

Bloom/Bloom.h
```diff
@@ -165,7 +166,7 @@ namespace Bloom {
 		// bloom filter bits
 
 		size_t bits = endBitPos - startBitPos + 1;
-		size_t bytes = (bits + 7) / 8;
+		uint64_t bytes = (bits + 7) / 8;
```
sjackman: Use `size_t`.

Common/StringUtil.h
```diff
@@ -106,7 +107,7 @@ bool isReadNamePair(const std::string& name1, const std::string& name2)
 	return false;
 }
 
-static inline size_t SIToBytes(std::istringstream& iss)
+static inline uint64_t SIToBytes(std::istringstream& iss)
```
sjackman: Use `uint64_t`.

Common/StringUtil.h
```diff
@@ -133,19 +134,19 @@ static inline size_t SIToBytes(std::istringstream& iss)
 
 	switch(tolower(units[0])) {
 		case 'k':
-			size *= (size_t)1<<10; break;
+			size *= (uint64_t)1<<10; break;
```
sjackman: Use `uint64_t`.

Common/StringUtil.h
```diff
@@ -133,19 +134,19 @@ static inline size_t SIToBytes(std::istringstream& iss)
...
 		case 'm':
-			size *= (size_t)1<<20; break;
+			size *= (uint64_t)1<<20; break;
 		case 'g':
```
sjackman: Use `uint64_t`.

Common/StringUtil.h
```diff
@@ -133,19 +134,19 @@ static inline size_t SIToBytes(std::istringstream& iss)
 
 	switch(tolower(units[0])) {
 		case 'k':
-			size *= (size_t)1<<10; break;
+			size *= (uint64_t)1<<10; break;
 		case 'm':
-			size *= (size_t)1<<20; break;
+			size *= (uint64_t)1<<20; break;
```
sjackman: Use `uint64_t`.

Common/StringUtil.h
```diff
@@ -133,19 +134,19 @@ static inline size_t SIToBytes(std::istringstream& iss)
...
 		case 'g':
-			size *= (size_t)1<<30; break;
+			size *= (uint64_t)1<<30; break;
```
sjackman: Use `uint64_t`.

Common/StringUtil.h
```diff
@@ -133,19 +134,19 @@ static inline size_t SIToBytes(std::istringstream& iss)
...
 		case 't':
-			size *= (size_t)1<<40; break;
+			size *= (uint64_t)1<<40; break;
```
sjackman: Use `uint64_t` and change `1<<40` to `1ULL << 40`.

Common/StringUtil.h
```diff
@@ -133,19 +134,19 @@ static inline size_t SIToBytes(std::istringstream& iss)
...
 		default:
 			iss.setstate(std::ios::failbit);
 			return 0;
 	}
 
-	return (size_t)ceil(size);
+	return (uint64_t)ceil(size);
```
sjackman: Use `uint64_t`.

benvvalk:
Thanks for your help, @mcirsta and @sjackman.  The Bloom.h code is mine, and I understand now why I need to be more careful about my use of unsigned long / uint64_t / size_t.   I will fix the code.

The purpose of the variable is to store the byte size of Bloom filters in memory, so size_t  is the most appropriate choice.

mcirsta:
Thanks, the truth is I'm fine with size_t too if it's large enough for the purpose, and ultimately you know the code better than I do , the changes proposed were just a way to fix some compile errors by someone who has no idea what the code does.
 
 This has been a good learning experience for me too as some things are not really obvious as far as C/C++ types go.

 The final decision on what to merge is yours and I've no problem with that.

sjackman:
@mcirsta Has this issue been resolved, that is, does ABySS compile cleanly on 32 bit architectures?

PR #73, opened 2015-06-10
Documentation fixes.

nsoranzo:
See commits for details.

It may be of your interest that I'm working on an updated Galaxy wrapper for ABySS (for abyss-pe in particular) which will go first to https://github.com/galaxyproject/tools-iuc/tree/master/tools/abyss and then on the Galaxy Tool Shed.


benvvalk:
Neat!

README.md
```diff
@@ -355,8 +355,8 @@ Parameters of the driver script, `abyss-pe`
  * `b`: maximum length of a bubble (bp) [`10000`]
  * `c`: minimum mean k-mer coverage of a unitig [`sqrt(median)`]
  * `d`: allowable error of a distance estimate (bp) [`6`]
- * `e`: minimum erosion k-mer coverage [`sqrt(median)`]
- * `E`: minimum erosion k-mer coverage per strand [`1`]
+ * `e`: minimum erosion k-mer coverage [`round(sqrt(median))`]
+ * `E`: minimum erosion k-mer coverage per strand
```
sjackman: I believe the default of the `E` parameter is `1`. No?

README.md
```diff
@@ -368,7 +368,7 @@ Parameters of the driver script, `abyss-pe`
  * `q`: minimum base quality [`3`]
  * `s`: minimum unitig size required for building contigs (bp) [`200`]
  * `S`: minimum contig size required for building scaffolds (bp) [`s`]
- * `t`: minimum tip size (bp) [`2k`]
+ * `t`: maximum length of dangling edges to trim [`k`]
```
sjackman: I prefer the term "blunt contig" (or "tip") to "dangling edge". In ABySS parlance, *k*-mers or contigs are vertices and sequence overlaps are edges.

sjackman: The question of `2*k` vs `k` is a matter of units. `k` *k*-mers are pruned, which is `2*k-1` bp. I'm not sure which unit is used by ABySS.

doc/abyss-pe.1
```diff
@@ -58,10 +58,10 @@ minimum mean k-mer coverage of a unitig [sqrt(median)]
 allowable error of a distance estimate (bp) [6]
 .TP
 .B e
-minimum erosion k-mer coverage [sqrt(median)]
+minimum erosion k-mer coverage [round(sqrt(median))]
 .TP
 .B E
-minimum erosion k-mer coverage per strand [1]
+minimum erosion k-mer coverage per strand
```
sjackman: What is the default value of `E`?

doc/abyss-pe.1
```diff
@@ -114,7 +114,7 @@ Assumes that the first read in a read pair is reveresed WRT the
 transcripts sequenced.
 .TP
 .B t
-minimum tip size (bp) [2k]
+maximum length of dangling edges to trim [k]
```
sjackman: See above.

sjackman:
Thanks for the documentation fixes, Nicola. This is fantastic!

sjackman:
Ben, please fix this Travis build error.
https://travis-ci.org/bcgsc/abyss/jobs/66208437#L995-L997
```
clang++ -Wall -Wextra -Wno-unknown-pragmas -g -O2  -L/usr/lib/openmpi/lib    -o PairedDBG_Dinuc PairedDBG/PairedDBG_Dinuc-DinucTest.o ../PairedDBG/libpaireddbg.a ../DataLayer/libdatalayer.a ../Common/libcommon.a ../lib/gtest-1.7.0/libgtest_main.a -ldl -lm 
../lib/gtest-1.7.0/libgtest_main.a(libgtest_main_a-gtest-all.o): In function `~ThreadLocal':
/home/travis/build/bcgsc/abyss/lib/gtest-1.7.0/../../lib/gtest-1.7.0/include/gtest/internal/gtest-port.h:1482: undefined reference to `pthread_getspecific'
```

README.md
nsoranzo: It's `0` or `1`: https://github.com/bcgsc/abyss/blob/master/Assembly/CoverageAlgorithm.h#L101

README.md
nsoranzo: I don't know the correct terminology, but it should be consistent with the one of `ABYSS --help` (which I copied here and below). I will `s/dangling edges/blunt contigs/` here, in `Assembly/Options.cc` and `doc/abyss-pe.1`, ok?

Regarding the default value, I assume it means the if not specified, the value of `k` will be used: https://github.com/bcgsc/abyss/blob/master/Assembly/Options.cc#L343

benvvalk:
@sjackman Okay, working on it (Travis error).

README.md
sjackman: Sounds good.

Unfortunately it looks as though `trimLen` is measured in *k*-mers and `bubbleLen` is measured in bp. Oi. That's not great.

README.md
sjackman: Ah, yes. Would you be up for documenting that as well?
`minimum erosion k-mer coverage per strand [1 if c > 2 else 0]`

nsoranzo: Sure, will do.

nsoranzo: Shouldn't it be `[1 if sqrt(median) > 2 else 0]`?

sjackman: Ah, yes. You're right.

nsoranzo:
@sjackman Fixed per your comments, thanks for reviewing!

nsoranzo:
I've added another doc fix just now.

sjackman:
@benvvalk Thanks for fixing the Travis build error.

doc/abyss-pe.1
```diff
@@ -49,7 +49,7 @@ files containing single-end reads
 maximum number of branches of a bubble [2]
 .TP
 .B b
-maximum length of a bubble (bp) [10000]
+maximum length of a bubble (bp) [3*k]
```
sjackman: ABySS has two bubble popping stages. One in `ABYSS` and one in `PopBubbles`. The value of the abyss-pe `b` parameter is passed to [both stages](https://github.com/bcgsc/abyss/blob/master/bin/abyss-pe#L150-L153). The first stage handles only equal-length bubbles (no indels), does not check sequence similarity (except that the first and last `k-1 bp` must be identical, because de Bruijn graph), and [the default value is `3*k bp`](https://github.com/bcgsc/abyss/blob/master/Assembly/Options.cc#L346). The second stages handles arbitrary bubbles (indels too), requires a minimum sequence similarity of `p` ([default `0.9`](https://github.com/bcgsc/abyss/blob/master/bin/abyss-pe#L191)), and [the default value is `10000 bp`](https://github.com/bcgsc/abyss/blob/master/PopBubbles/PopBubbles.cpp#L99).

So, I think here it's more accurate to say that the maximum bubble length is `10000`, but it's complicated.

sjackman: Let's just document the truth and say
`maximum length of a bubble (bp) [3*k for ABYSS and 10000 for PopBubbles]`
Done.

nsoranzo: Thanks @sjackman for the detailed explanation! I think it is very bad that running the program without specifying a value for an option may give different results from specifying the default value for it.

Maybe an interim solution may be to pass the value of `b` only to `PopBubbles` (default `10000`) and leave `ABYSS` to always run with its default of `3*k`, what do you think?

sjackman:
> Thanks @sjackman for the detailed explanation! I think it is very bad that running the program without specifying a value for an option may give different results from specifying the default value for it.

I agree 100%. We should document that the default value for `b` is the empty string. That is: `b=`

> Maybe an interim solution may be to pass the value of b only to PopBubbles (default 10000) and leave ABYSS to always run with its default of 3*k, what do you think?

I would be okay with this suggestion, except that it's very important that setting `b=0` disable bubble popping altogether. I'm inclined to leave it in its current nasty state, for consistency of behaviour with previous releases, and then disable bubble popping at the ABYSS stage in the next release.

nsoranzo:
So `maximum length of a bubble (bp) []` ?

nsoranzo:
FYI, here's the new the Galaxy wrapper: https://github.com/galaxyproject/tools-iuc/pull/192

benvvalk:
@nsoranzo This is great.  Thanks for the fixes!  Regarding the documentation of the `b` param, can we please go with Shaun's suggestion?:
```
maximum length of a bubble (bp) [3*k for ABYSS and 10000 for PopBubbles]
```
IMO, this is the best solution for the short term.  As Shaun said, in the next release we will try to remove the ugliness of having two bubble popping stages.

sjackman:
I'd suggest
```
maximum length of a bubble (bp) [""] (the empty string)
ABySS has two bubble popping stages. The default limits are 3*k bp for ABYSS and 10000 bp for PopBubbles.
```

sjackman:
Thanks for all this detailed work, Nicola. It's very helpful to document reality accurately. And thanks for the Galaxy formula! I was wondering what your motivation was for this work.

benvvalk:
@sjackman sounds good to me :-)

nsoranzo:
Should be ready to merge now, thanks for all the comments!

sjackman:
Looks good to me. Nicola, can you squash these seven commits into a single commit?

nsoranzo:
Rebased and squashed in 2 commits.

sjackman:
Thanks! @benvvalk I'm happy to merge this.

benvvalk:
@sjackman sure, go ahead.  Thanks for reviewing the changes, too!

sjackman:
Thanks for your contribution, Nicola, and thanks for the Galaxy wrapper.

PR #76, opened 2015-06-30
use more portable /usr/bin/env perl for scripts

wookietreiber:
This is useful, e.g. if your perl is not installed in /usr/bin or you want to give a specific perl precendence over the one installed in /usr/bin.


sjackman:
:+1:

benvvalk:
Can't argue with that.  Thanks!

Issue #79, opened 2015-08-25
abyss-sealer v1.9.0 fails to run to completion


dparks1134:
Hello,

I have a number of assembled metagenomes I am running through abyss-sealer. The majority of these finish without issue in ~6 hours. However, a small subset never appear to complete. I have allowed these to run for 100+ hours at this point without completion. The offending metagenomes are ~1/3 lane of Illumina, and are similar in both number of assembled contigs and raw reads as the metagenomes that finish. Furthermore, I can see that abyss-sealer is running successfully for larger k values (k>=50), but getting 'hung' on k=40. I suspect there is an infinite loop in the code.

Have you experienced this problem? Do you have any suggestions on why some metagenomes may take excessive computation time for specific k values?

Cheers,
Donovn

benvvalk:
Hi Donovan,

Yes, I have seen that behaviour unfortunately.

I believe it is not an infinite loop but searches in really dense regions of the graph that cause problems.  At the moment, all that can be done is to adjust parameters to get a more reasonable runtime. The parameters that you could change are (in order of biggest effect to smallest effect):
* `--kmer`,`-k` (increase to make it run faster (as you have already observed))
* `--max-frag`, `-F` (limits search depth, i.e. maximum gap size; decrease to make it run faster)
* `--max-branches`,`-B` (search is aborted when breadth of search frontier exceeds this limit; decrease to make it run faster)

I do believe that this problem is fixable by making adjustments to the search algorithm, but a fix is not imminent.

benvvalk:
Multithreading (`-j`) helps a great deal too, of course, but you are probably already using that.

dparks1134:
Thanks. Changing the maximum gap size (<code>-F</code>) had a huge impact on the running time.

benvvalk:
Glad to hear!

Just a note: I wasn't precise in my response above: -F is actually max_gap_size + 2*flank_length, where flank_length (-L) is 100 bp by default.

dparks1134:
Thanks! Good to know.

sjackman:
Hi, Ben. As this issue is blocking #80, I'm going to reopen it. What can we do to limit the runtime of Sealer to something reasonable even for worst-case data?

benvvalk:
Fixed in [ABySS-2.1.2](https://github.com/bcgsc/abyss/releases/tag/2.1.2).

PR #80, opened 2015-08-27
add in abyss-sealer to abyss-pe run

macmanes:
not sure if leaving out sealer in the 'normal' abyss-pe is intended or an oversight.. Pull request adds this in if it's a mistake..

### DIFF
--- a/bin/abyss-pe
+++ b/bin/abyss-pe
@@ -357,7 +357,7 @@ ifneq ($(in),)
 default: contigs contigs-graph
 endif
 ifneq ($(mp),)
-default: scaffolds scaffolds-graph
+default: scaffolds scaffolds-graph seal-scaffolds
 ifneq ($(long),)
 default: long-scaffs long-scaffs-graph
 endif
@@ -644,7 +644,7 @@ endif
 sealer_ks?=-k90 -k80 -k70 -k60 -k50 -k40 -k30
 
 %-8_scaffold.fa: %-8.fa
-	abyss-sealer -v -j$j --print-flanks -o$*-8 -S$< $(sealer_ks) $(SEALER_OPTIONS) $(in) $(se)
+	abyss-sealer -v -j$j --print-flanks -o $*-8 -S $< $(sealer_ks) $(SEALER_OPTIONS) $(in) $(se)
 
 %-scaffolds-sealed.fa: %-8_scaffold.fa
 	ln -s $< $@


benvvalk:
Thanks for the patch, Matt.

Cross-posting my response from the ABySS Google Group:

--- BEGIN QUOTE ---
abyss-sealer is disabled by default because it sometimes has unpredictably long run times at low k values.  See for example: https://github.com/bcgsc/abyss/issues/79

That said, the tool is highly useful and generally produces good results, but sometimes you have to fiddle with the parameters to get reasonable execution times.

When we sort the run time issue out, we will make it part of the pipeline by default.
--- END QUOTE ---

sjackman:
Hi, Ben. I believe the Sealer search is limited by both breadth and depth. One suggestion was to add an overall number-of-vertices-visited limit. If every search hit that limit, it could still take quite a long time to complete. Any thoughts?

bin/abyss-pe
```diff
@@ -644,7 +644,7 @@ endif
 sealer_ks?=-k90 -k80 -k70 -k60 -k50 -k40 -k30
 
 %-8_scaffold.fa: %-8.fa
-	abyss-sealer -v -j$j --print-flanks -o$*-8 -S$< $(sealer_ks) $(SEALER_OPTIONS) $(in) $(se)
+	abyss-sealer -v -j$j --print-flanks -o $*-8 -S $< $(sealer_ks) $(SEALER_OPTIONS) $(in) $(se)
```
sjackman: @macmanes Please don't include white-space-only changes.

sjackman:
Blocked by #79.

bin/abyss-pe
macmanes: I think the issue with the whitespace changes was tat the options did not parse correctly on my machine without them... It's been a month, so my memory is not crystal clear about this point however.

sjackman: That shouldn't be the case and is a bug if so.

benvvalk:
@sjackman: Yes, the vertices limit is still a good idea.  Just haven't gotten around to it.

sjackman:
@benvvalk Should ABySS run Sealer by default?

benvvalk:
@sjackman Yes, this is still a highly desirable quality-of-life improvement, as is the related performance fix to Sealer and Konnector that I have put off for years now.

However, it is way below spruce-related work and ABySS 2.0 improvements on my priority list, both of which are going to keep me busy for the foreseeable future.

PR #81, opened 2015-08-27
Add PacBio scaffolding

macmanes:
New abyss-pe functionality which adds formal support for scaffolding with PacBio reads. 

Input pacbio sequences via `pacbio='name'`. Processed PB using transcriptome scaffolded data. 

`bwa mem -x pacbio` mapping for PB sequences
also lowered gap initiation and extension penalties in bwa mem mapping for transcriptome data. 

This results in a fairly sizable jump in NG50 and other length based metrics.

**ISSUES:**

Currently does not work when _just_ pacbio reads are used and not transcriptome + pacbio, as the pacbio scaffolding stuff builds off -10.fa. Simple enough to build in an `if` statement, but given this is an 'unsolicited' pull request that you may not accept, I didn't want to spend time building in this functionality.. 

There are many "Warning: malformatted CIGAR string" messages during `abyss-longdistseq`. This is an known issue and not sure if these warnings mean anything or should be suppressed.

### DIFF
--- a/bin/abyss-pe
+++ b/bin/abyss-pe
@@ -2,6 +2,7 @@
 # Run the ABySS assembler.
 # Written by Shaun Jackman <sjackman@bcgsc.ca> and
 # Anthony Raymond <traymond@bcgsc.ca>.
+#PacBio functionality written by Matthew MacManes <matthew.macmanes@unh.edu>
 
 # Set pipefail to require that all commands of a pipe must succeed.
 SHELL=/bin/bash -o pipefail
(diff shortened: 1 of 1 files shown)


macmanes:
any interest in adding this functionality?

sjackman:
Hi, Matt. Thanks for the patch. I definitely want the functionality of scaffolding with PacBio data in ABySS. My preference would be to merge this functionality into the `long` option. See #83 for example. We could still include a `pacbio` option for convenience, but it may just set the options `long` and `LONG_BWAMEM_OPTIONS`. Your thoughts, Ben?

macmanes:
I think it would be fine to leave it in the `long` option, so long as users could map pacbio reads with `-x pacbio` and whatever other long reads with `-x nanopore` or whatever for transcriptomes.

benvvalk:
Thanks for your contribution, Matt. :-)

Thanks, Shaun.  I agree that having `long` support multiple libraries and separate options for each library is the way to go (as suggested by you on #83).   That would address the issue with Matt's current patch requiring transcriptome data in order to do the long scaffolding.

PR #108, opened 2016-04-21
Fix BUGs of abyss-tofastq 

jiguanglizipao:
Fix BUG:abyss-tofastq cannot convert .fa to .fastq (EX. Use Dida)
Fix BUG:make --cat become default rather than -i (help info said --cat should be defult)

### DIFF
--- a/DataLayer/FastaReader.h
+++ b/DataLayer/FastaReader.h
@@ -168,11 +168,14 @@ struct FastqRecord : FastaRecord
 {
 	/** Quality */
 	std::string qual;
-
-	FastqRecord() { }
+	char default_qual;
+	FastqRecord()
+		:default_qual('2')
+	{
+	}
 	FastqRecord(const std::string& id, const std::string& comment,
-			const Sequence& seq, const std::string& qual)
-		: FastaRecord(id, comment, seq), qual(qual)
+			const Sequence& seq, const std::string& qual, const char _default_qual = '2')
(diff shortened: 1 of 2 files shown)


sjackman:
> Fix BUG:make --cat become default rather than -i (help info said --cat should be defult)

I believe `--cat` is the default. Please provide a test case if you're not seeing this behaviour.

sjackman:
> Fix BUG:abyss-tofastq cannot convert .fa to .fastq (EX. Use Dida)

It's by design the ABySS refuses to convert a FASTA to FASTQ and make up fake quality values. Whether it's good design or bad design can be debated.

If Dida can't read FASTA files, I'd suggest opening up a feature request issue asking @mohamadi to implement that feature. https://github.com/bcgsc/dida/issues/new

jiguanglizipao:
Sorry, cat is default after test again.
I think we can add a new parameter to this part. Do you think it is a new feature? This bug doesn't only happen in Dida, but also in every aligner using this part(need fastq). So although this patch may not be well designed, I still submit this pull request because this bug is critical.

jiguanglizipao:
If you think a  new parameter like "--force" is ok, I  will do that soon.

sjackman:
Do you know of other aligners other than DIDA that will only accept FASTQ. I'd rather fix this issue in DIDA. I work with the developer of DIDA, so I'll chat with him tomorrow.

Issue #75, opened 2015-06-23
Invalid LICENSE file 


kblin:
Hi folks,
The usual "I am not a lawyer" disclaimer applies for the following.
While I very much appreciate you distributing ABySS under a GPLv3 license, I don't think your "non-commercial" additional requirement results in a valid license. While of course the copyright holder gets to pick whatever license they like, if I were to distribute ABySS without making money of it, I could only distribute it under the GPL, as the GPL doesn't allow me to add any additional restrictions. Thus, the "non-commercial" part would be lost in the intermediate step, allowing all use under the GPLv3. I assume that's not what you had in mind when you selected this wording on the license.
Additionally, I don't think you're allowed to use the GPL license text in the current altered form.

Regards,
Kai

sjackman:
Thanks for you concern, Kai. It is a sticky point, and I do not know the answer for certain. My interpretation is below. I also am not a lawyer, and I also am not the copyright holder, and so I do not pick the license, though I do have input.

> I could only distribute it under the GPL, as the GPL doesn't allow me to add any additional restrictions. Thus, the "non-commercial" part would be lost in the intermediate step, allowing all use under the GPLv3.

The license gives you the right to redistribute the software. I do not believe that the recipient receiving a copy of the software from you also conveys to the recipient a license to run the software. That right may only be conveyed by the copyright holder.

> Additionally, I don't think you're allowed to use the GPL license text in the current altered form.

The GPL license text is not modified. The GPL license text starts at [line 24, "GNU GENERAL PUBLIC LICENSE"](https://github.com/bcgsc/abyss/blob/master/LICENSE#L24). Everything above that is preamble.

tjanez:
@sjackman, I'm afraid you reasoning is incorrect.

The [GPL FAQ](http://www.gnu.org/licenses/gpl-faq.html#NoMilitary) clearly states that one can not license code under the GPL and add a restriction that it can't be used for commercial purposes:
> **I'd like to license my code under the GPL, but I'd also like to make it clear that it can't be used for military and/or commercial uses. Can I do this? (#NoMilitary)**

>  No, because those two goals contradict each other. The GNU GPL is designed specifically to prevent the addition of further restrictions. GPLv3 allows a very limited set of them, in section 7, but any other added restriction can be removed by the user.

So, the software is actually not licensed under GPLv3. Rather, it uses a custom license that is similar to GPLv3 but removes the freedom to use the software for commercial purposes.

[...]

PR #124, opened 2016-09-06
Be specific about licensing terms in COPYRIGHT

ttriche:
This pull request is in response to a Twitter tempe^H^H^H^H^Hexchange which seemed overblown.  The most rational course of action would appear to be explicit licensing of components as currently stands in COPYRIGHT, with implicit licensing of additional components "compromised" by GPLv3.  However, the GPLv3-licensed components, along with sections 1, 7, and 10 of the GPLv3, render the point moot; either the entire project is GPLv3, or it needs to be relicensed and clean-roomed. 

It turns out that by specifically GPLv3 licensing (e.g.) `Common/*`, any code that happens to directly 
`#include` code from `Common/` is "infected" by the GPLv3, sections 7 and 10, which specifically disallow further restrictions save for very narrow exceptions (see https://www.gnu.org/licenses/gpl-faq.en.html#NoMilitary, which is of course not binding *per se*).  `Common/*` infects `Graph/*`, `Bloom/*`, and most branches of the code either directly or indirectly in this fashion, thus they are implicitly licensed under GPLv3 terms themselves (at least in their current form).

I can find nothing to *directly, explicitly* suggest GPLv3 "infection" of
```
autogen.sh bin ChangeLog circle.yml CITATION.bib CITATION.md configure.ac 
COPYRIGHT doc Dockerfile doxygen.conf IntegrationTest lib LICENSE m4 
Makefile.am Misc README.md
```
although section 1 of the GPLv3 would lump most of these in with "build scripts, etc".  It would be a good idea to spell out which items are explicitly GPLv3, LGPLv2.1, BSD 3-clause, and/or Expat licensed, which are explicitly sublicensed under GPLv3 for non-commercial use, and by omission, those implicitly infected by GPLv3 licensing at compile time in the current configuration.  At the very least this would clarify what was not already GPLv3 or otherwise OSS licensed at this time.

I am not a lawyer, nor am I licensed to practice law in any US state or Canadian province.  The only settled legal precedents for GPL "infection" occur in Belgian and German case law.  A Canadian judge might well rule differently.  Nonetheless, it cannot hurt to be explicit about licensing terms.  Furthermore, since nearly all of the code under GPLv3 derives from Shaun Jackman or other BCCA/BCGSC employees, there is absolutely no reason that it cannot be re-licensed *en bloc* to fall under a noncommercial license (all contributors must agree to a license change, but all contributors have presumably assigned rights to BCCA/BCGSC as a condition of employment).

This PR seeks solely to disambiguate licensing cross-contamination as it currently exists in ABySS.

### DIFF
--- a/COPYRIGHT
+++ b/COPYRIGHT
@@ -22,7 +22,7 @@ License: GPL-NC-3+
  .
  Debian may redistribute this software package.
 
-Files: autogen.sh bin/* ChangeLog circle.yml CITATION.bib CITATION.md configure.ac COPYRIGHT doc/* Dockerfile doxygen.conf IntegrationTest/* lib/* LICENSE m4/* Makefile.am Misc/* README.md
+Files: autogen.sh bin/* ChangeLog circle.yml CITATION.bib CITATION.md configure.ac COPYRIGHT doc/* Dockerfile doxygen.conf IntegrationTest/* LICENSE m4/* Makefile.am Misc/* README.md
 Copyright: Copyright 2014 Canada's Michael Smith Genome Sciences Centre
 License: GPL-NC-3+
 
--- a/COPYRIGHT
+++ b/COPYRIGHT
@@ -22,7 +22,7 @@ License: GPL-NC-3+
  .
  Debian may redistribute this software package.
 
-Files: *
+Files: autogen.sh bin/* ChangeLog circle.yml CITATION.bib CITATION.md configure.ac COPYRIGHT doc/* Dockerfile doxygen.conf IntegrationTest/* lib/* LICENSE m4/* Makefile.am Misc/* README.md
 Copyright: Copyright 2014 Canada's Michael Smith Genome Sciences Centre
 License: GPL-NC-3+
 


kblin:
I would like to point out again (as I did in #75) that there is no such thing as "GPLv3 for non-commercial use".
Of course the authors are free to change the license on all parts they own copyright to and that are not derived works of a copyleft-licensed piece of code, but at that point it's not GPLv3 anymore.

jmchilton:
My "temper tantrum" (i.e. desire to do open source in good faith) of an issue with the license isn't the contamination aspect - though I'm glad those are being reviewed and considered. It is that this license `GPL-NC-3+` is nonsensical: 

```
License: GPL-NC-3+
 You may use, redistribute and modify this software for non-commercial
 purposes under the terms of the GNU General Public License
```

Which says if I don't have a commercial purpose (and I don't) I can redistribute it under the terms of the GPL. If I can redistribute it under the GPL - I explicitly cannot add a non-commercial condition ... or in other words ... I can redistribute freely and unencumbered as allowed (and legally required) by the license.

The right solution is just to drop the -NC- part of this license all together, if this is not done then someone really should fork and drop it - because that is allowable and because the current state of affairs is not legal or ethical open source. Everyone who wrote the GPL and LGPL modules that forced this project to be GPL in the first place devoted their time to those projects with the understanding that derived works would be copy-left open source - and this is not.

kblin:
On a more technical side, lib/gtest-1.7.0 is copyright by Google, so can't be re-licensed by the abyss authors, so your `lib/*` in the pull request is incorrect.

ttriche:
fixed that.  

@jmchilton I *think* @sjackman is the sole author of all those GPL'ed libraries, hence in principle he could just re-license all of it going forward under something else.  And, yeah, that was a lot of effort, so maybe give him the benefit of the doubt?  

@kblin would be justified in pointing out "but issue #75 was opened many months ago!", nonetheless, this is a lot of code mostly written by one person and I am hopeful that this can be resolved sensibly.

ttriche:
What's really odd (to my mind) is that a BCCA license exists, and is used for other BCGSC projects, rather than an unusual "rewrite" of the GPLv3.  I don't see any reason why a person cannot seek remuneration if a commercial enterprise profits from their hard work; however, this edited GPLv3 is probably not the way to go about that.  The BCCA license actually seems to make sense.

sjackman:
I am not the copyright holder of ABySS and cannot unilaterally change the license of ABySS. The copyright holder is the BC Cancer Agency. I can make recommendations, and I can request that the license be changed.

> Everyone who wrote the GPL and LGPL modules that forced this project to be GPL in the first place devoted their time to those projects with the understanding that derived works would be copy-left open source - and this is not

You seem to be confused here. There aren't any such libraries, and nothing forced anyone to change the license of ABySS to GPL. All of the GPL-NC-3+/GPL-3+ code in ABySS was written by employees of the BC Cancer Agency, primarily Jared Simpson (@jts), myself (@sjackman) and Ben Vandervalk (@benvvalk), and the copyright is owned by the BCCA.

The license of ABySS was originally the BCCA Academic License, which is a restrictive and non-commerical license. It was changed to GPL-NC-3+, because it is more permissive than the BCCA Academic License, easier to understand and deemed better than the original BCCA Academic License, which is not good.

The code of certain modules (`Common DataLayer DistanceEst FMIndex Map ParseAligns`) was changed from GPL-NC-3+ to GPL-3+, because Jared Simpson (@jts) asked me if he could use those modules in his program SGA, I asked the BCCA technology transfer office, and they said yes, and so those modules were relicensed.

Since then I have learned that it's effectively impossible to add a non-commercial restriction to the GPL-3+. I will e-mail the technology transfer office to ask whether we can change the license of ABySS to GPL-3+.

ttriche:
Cool!  Thanks for doing this!  I did not know (although I suspected) that
it was you who requested GPLv3 for those modules.  I noticed when reading
the code that you and Jared Simpson appeared to be the authors of almost
everything under GPL or without an assigned BCCA license.

Hopefully the tech transfer office will come to the same conclusion (that
superimposing a noncommercial restriction on code that uses GPLv3 modules
is legally intractable).  I learned a good deal while looking through the
code, it's obvious what was written early on vs. later due to the evolution
of your style.  It would be a shame if others could not (due to clean-room
clauses and that sort of thing).


--t

sjackman:
> it's obvious what was written early on vs. later due to the evolution
of your style

😳 I've learned a lot in 8 years. The biggest influence on my style was reading [The Boost Graph Library: User Guide and Reference Manual](https://www.amazon.ca/Boost-Graph-Library-Reference-Manual/dp/0201729148), which fully converted me from object-oriented to generic programming. It's a great book for learning more generic programming using graphs as the use-case.

sjackman:
We've met with the technology transfer office, and the license of ABySS will be changed to GPL-3! Ben @benvvalk will be making a release of ABySS 2.0.1 shortly that changes the license to GPL-3.

PR #168, opened 2017-09-28
Documentation: remove "disabled" from params

zkamvar:
When I saw that `G` was "disabled", I had interpreted this to mean that it was unavailable for me to modify, but it looks like this is simply an option without a default. 

To be more consistent with options `x` and `K` (which also do not have defaults), I have removed `[disabled]` from `G` and `v`. I also sprinkled in a few backticks for consistency.


sjackman:
Excellent. Thank you for your first contribution, Zhian! :heart:

benvvalk:
Ditto.  Thanks, @zkamvar! (and @sjackman)

PR #174, opened 2017-11-22
fix OSX/clang linking issue due to `&std::string::length`

mbargull:
Compiling/linking on OSX with `clang` and `libc++` fails due to using a pointer to member function `std::string::length` which apparently isn't exported for `libc++`, see:
http://lists.llvm.org/pipermail/cfe-dev/2013-March/028407.html

With C++11 and onward a lambda function could be used. Since this project doesn't appear use C++11, a functor is used instead.

### DIFF
--- a/MergePaths/PathConsensus.cpp
+++ b/MergePaths/PathConsensus.cpp
@@ -533,6 +533,13 @@ static ContigPath alignPair(const Graph& g,
 	return path;
 }
 
+template <typename Seq>
+struct GetLength {
+    size_t operator()(const Seq& sequence) {
+        return sequence.length();
+    }
+};
+
 /* Resolve ambiguous region using multiple alignment of all paths in
  * `solutions'.
  */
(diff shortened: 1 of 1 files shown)


mbargull:
Build failure with `clang` is unrelated:
```
../Graph/AllPathsSearch.h:30:5: error: unused typedef 'boost_concept_check30'
```

benvvalk:
Thank you, @mbargull!

Giving it a quick test on Linux before I merge.

benvvalk:
Compiled fine with gcc-4.4/CentOS5 and gcc-5.3/CentOS7 and all tests pass.

Thanks again!

sjackman:
> std::string::length which apparently isn't exported for libc++

How… odd. For the record, the error is…

```
Undefined symbols for architecture x86_64:
  "std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::length() const", referenced from:
      _main in assignment-F0ysIg.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Thanks, Marcel!

mbargull:
Odd indeed! Thanks for merging this so promptly!
BTW: This patch enabled us to build versions `1.5.2`, `1.9.0`, and `2.0.2` on OSX for the [Bioconda](https://bioconda.github.io) channel: https://anaconda.org/bioconda/abyss/files.

PR #191, opened 2018-02-08
Clarify abyss-fixmate output messages

mmokrejs:


ParseAligns/abyss-fixmate.cc
```diff
@@ -327,7 +334,7 @@ static void printHistogramStats(Histogram h)
 		"n: " << h.size() << " "
 		"min: " << h.minimum() << " "
 		"max: " << h.maximum() << " "
-		"ignored: " << n_orig - h.size() << '\n'
+		"ignored as outliers: " << n_orig - h.size() << '\n'
```
sjackman: I'd suggest `trimmed`

ParseAligns/abyss-fixmate.cc
```diff
@@ -488,7 +495,7 @@ int main(int argc, char* const* argv)
 
 	if (alignments.size() == sum) {
 		cerr << PROGRAM ": error: All reads are mateless. This "
-			"can happen when first and second read IDs do not match."
+			"can happen when first and second read IDs are not in same order."
```
sjackman: Order does not matter. Please revert this change.

ParseAligns/abyss-fixmate.cc
```diff
@@ -452,14 +459,14 @@ int main(int argc, char* const* argv)
...
-		"Different  " << percent(stats.numDifferent, sum) << "\n"
-		"Total      " << sum << endl;
+		"Mateless (read has no mate or input ordering is wrong) " << percent(alignments.size(), sum) << "\n"
+		"Unaligned (both reads unaligned)                       " << percent(stats.bothUnaligned, sum) << "\n"
+		"Singleton (one read aligned, one unaligned)            " << percent(stats.oneUnaligned, sum) << "\n"
+		"FR (forward-reverse, commonly paired-end reads)        " << percent(numFR, sum) << "\n"
+		"RF (reverse-forward, true mate-pair reads)             " << percent(numRF, sum) << "\n"
+		"FF (unsupported by abyss)                              " << percent(stats.numFF, sum) << "\n"
+		"Different (each read maps to a different contig)       " << percent(stats.numDifferent, sum) << "\n"
+		"Total                                                  " << sum << endl;
```
sjackman: I'd prefer to move this information to documentation. The Wiki is publicly editable. Feel free to create a new page at https://github.com/bcgsc/abyss/wiki

ParseAligns/abyss-fixmate.cc
```diff
@@ -34,7 +34,14 @@ static const char USAGE_MESSAGE[] =
 "Usage: " PROGRAM " [OPTION]... [FILE]...\n"
 "Write read pairs that map to the same contig to the file SAME.\n"
 "Write read pairs that map to different contigs to stdout.\n"
-"Alignments may be in FILE(s) or standard input.\n"
+"Alignments may be in SAM-formatted FILE(s) or standard input.\n"
```
sjackman: ✅

ParseAligns/abyss-fixmate.cc
```diff
@@ -34,7 +34,14 @@ static const char USAGE_MESSAGE[] =
...
+"\n"
+"The distribution of fragment sizes calculated from read pairs\n"
+"that map to the same contig is first trimmed (0.0001 from each\n"
+"end) to remove outliers before calculating the mean and standard\n"
+"deviation. Read pairs that map to the same contig are not useful\n"
+"for scaffolding, other than for this purpose of calculating the\n"
+"empirical distribution of fragment sizes.\n"
```
sjackman: I'd prefer to move this information to documentation. The Wiki is publicly editable. Feel free to create a new page for `abyss-fixmate` at https://github.com/bcgsc/abyss/wiki

ParseAligns/abyss-fixmate.cc
mmokrejs: But this is not really trimming. Trimming in my opinion means removing=altering e.g. an adapter sequece from input reads. Provided it is critical for end-user to understand what is happening with input data (reads/k-mers generally speaking) I think trimming would be unfortunate. I understand you mean "trimming from the distibution" but it is a long shot to understand would there be just 'trimming'. In my opinion `ignored as outliers` is clear to understand, does not require to rename the internal variable `ignored`. Abyss-map just omits those read pairs from further work.

You are the gate-keeper here but ... just my 2c.

ParseAligns/abyss-fixmate.cc
mmokrejs: Done in 4348e7a3cc99e791370555ffff629395d313e258

ParseAligns/abyss-fixmate.cc
mmokrejs: Umm, that was the main reason I wanted to clarify the *log file*. People want to copy&paste relevant info from the logs and the columns should be self-explanatory. You are free to reject me. I do not say including this text in addition into the wiki pages too would be that much work but I haven't heard of those pages before. Abyss has disparate documentation scattered in runtime help texts, all the README.md files, I would rather improve clarify of the log which are the primary source of the information.

ParseAligns/abyss-fixmate.cc
mmokrejs: You know, `abys-fixmate` sounds like 'fixing' mate information in an input stream, and outputting a stream with fixed content out. It is actually not what this utility is doing. It splits the input into pairs of reads which mapped to same reference contig from those which mapped to different contigs. Instead of proposing renaming of the executable (which would break 3rd-party apps) I at least wanted to make clear what this tool really does.

ParseAligns/abyss-fixmate.cc
sjackman: Trimmed mean is a statistical term. See https://en.wikipedia.org/wiki/Trimmed_estimator
A synonym is truncated mean. https://en.wikipedia.org/wiki/Truncated_mean
I'd be okay with `truncated`.

They're not exactly outliers, which is usually defined as x > mean + n * sigma or x > median + n IQR. That'd probably be a better trimming method than the fixed proportion of trimming that we're doing currently.

ParseAligns/abyss-fixmate.cc
sjackman: There's a limit to how much documentation you can include in the output of the program itself. I'd prefer to keep the output terse and improve the documentation, which there's more room for explanation. I'd like each tool of ABySS to have a page of documentation, like Sealer does, to describe its output.

ParseAligns/abyss-fixmate.cc
sjackman: When `--hist` is not specified, `abyss-fixmate` does the identical job as `samtools fixmate`, from which its name is derived. When `--hist` is specified, it removes reads that map to the same contig.

sjackman: That information could be added to the description of the `--hist` option.

PR #193, opened 2018-02-15
Add .github/stale.yml

sjackman:
Enable Stalebot to close stale issues.
See https://github.com/probot/stale
and https://github.com/apps/stale


benvvalk:
Nice.  Thank you, Shaun.

PR #195, opened 2018-02-16
CircleCI: Use clang

sjackman:
A subsequent PR can test with both GCC and Clang.


sjackman:
CircleCI passes with Clang 3.8.0. It appears that TravisCI is using Clang 5.0.0. We should try that version of Clang on CircleCi.
https://travis-ci.org/bcgsc/abyss/jobs/342187716#L456

Issue #205, opened 2018-03-21
abyss-2.0.3: make: command: Command not found


mmokrejs:
Hi,
  I am testing openmpi and abyss compiled by our sysadmin. I do have make available but nevertheless, the following command with abyss-2.0.3 shows an error which is silently ignored (bad). Execution of the very same command with abyss-2.0.2 does not result in `make: command: Command not found` message.

```
$ make
make: *** No targets specified and no makefile found.  Stop.
$ abyss-pe v=-v np="$threads" j="$threads" k="${kmersize}" name=test_"${kmersize}" pe-bam mp-bam default lib='9998' 9998=9998.fasta
make: command: Command not found
mpirun --verbose  -np 24 ABYSS-P -k160 -q3 -v   --coverage-hist=coverage.hist -s test_160-bubbles.fa  -o test_160-1.fa 9998.fasta 
ABySS 2.0.3
ABYSS-P -k160 -q3 -v --coverage-hist=coverage.hist -s test_160-bubbles.fa -o test_160-1.fa 9998.fasta
Running on 24 processors
1: Running on host login3
3: Running on host login3
5: Running on host login3
7: Running on host login3
9: Running on host login3
11: Running on host login3
13: Running on host login3
15: Running on host login3
17: Running on host login3
19: Running on host login3
21: Running on host login3
23: Running on host login3
0: Running on host login3
2: Running on host login3
4: Running on host login3
8: Running on host login3
10: Running on host login3
12: Running on host login3
16: Running on host login3
18: Running on host login3
20: Running on host login3
6: Running on host login3
14: Running on host login3
22: Running on host login3
```

```
$ make -d -f /apps/all/ABySS/2.0.3-foss-2018a/bin/abyss-pe v=-vvvvvvvvvvvvv np="$threads" j="$threads" k="${kmersize}" name=test_"${kmersize}" pe-bam mp-bam default lib='9998' 9998=9998.fasta
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for x86_64-redhat-linux-gnu
Reading makefiles...
Reading makefile `/apps/all/ABySS/2.0.3-foss-2018a/bin/abyss-pe'...
make: command: Command not found
Updating makefiles....
[...]

PR #206, opened 2018-03-22
abyss-pe: Fix make: command: Command not found

sjackman:
GNU make does not appear to launch a shell when SHELL=/bin/sh.

Fixes https://github.com/bcgsc/abyss/issues/205


benvvalk:
Thanks, Shaun!

PR #221, opened 2018-05-01
Updated README.md [Before Starting an Assembly]

kkapuria3:
If root directory is less than size of genome then first set TMPDIR before you execute ABySS,

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,9 +192,9 @@ installed ABySS in `/opt/abyss`, add `/opt/abyss/bin` to your `PATH`:
 Before starting an assembly
 ===========================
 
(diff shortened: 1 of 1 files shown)


sjackman:
Thanks for the documentation fix, Karan! My I suggest the following wording.

ABySS stores temporary files in `TMPDIR`, which is `/tmp` by default on most systems. If your default temporary disk volume is too small, set `TMPDIR` to a larger volume, such as `/var/tmp` or your home directory.

```sh
export TMPDIR=/var/tmp
```

kkapuria3:
reads perfect.

sjackman:
Thanks for your first contribution to the ABySS documentation, Karan!

PR #242, opened 2018-07-23
Add kseq.h lib to FastaReader

schutzekatze:
Some notes:
- kseq.h requires zlib, so I had to add it to linker flags to several Makefile.am-s.
- Most of the time dealing with fasta/fastq files is spent unzipping them (if they're zipped), but because of zlib, kseq.h allows for extracting while reading, instead of extracting the whole file, which is slightly faster.
- Because kseq.h doesn't have peek(), there's a filename based check to see whether the file is fasta/fastq, otherwise it falls back to original implementation.

### DIFF
--- a/ABYSS/Makefile.am
+++ b/ABYSS/Makefile.am
@@ -7,6 +7,6 @@ ABYSS_LDADD = \
 	$(SQLITE_LIBS) \
 	$(top_builddir)/Assembly/libassembly.a \
 	$(top_builddir)/DataLayer/libdatalayer.a \
-	$(top_builddir)/Common/libcommon.a
+	$(top_builddir)/Common/libcommon.a -lz
 
 ABYSS_SOURCES = abyss.cc
--- a/AdjList/Makefile.am
+++ b/AdjList/Makefile.am
@@ -10,7 +10,7 @@ endif
 
 AdjList_LDADD = \
 	$(top_builddir)/DataLayer/libdatalayer.a \
-	$(top_builddir)/Common/libcommon.a
+	$(top_builddir)/Common/libcommon.a -lz
 
 if PAIRED_DBG
 AdjList_LDADD += \
@@ -19,7 +19,7 @@ endif
 
 AdjList_LDADD += \
 	$(top_builddir)/DataBase/libdb.a \
-	$(SQLITE_LIBS)
+	$(SQLITE_LIBS) -lz
 
 AdjList_SOURCES = \
 	AdjList.cpp
--- a/Align/Makefile.am
+++ b/Align/Makefile.am
@@ -18,7 +18,7 @@ abyss_align_CXXFLAGS = $(AM_CXXFLAGS) $(OPENMP_CXXFLAGS)
 abyss_align_LDADD = $(builddir)/libalign.a \
 	$(top_builddir)/dialign/libdialign.a \
 	$(top_builddir)/DataLayer/libdatalayer.a \
-	$(top_builddir)/Common/libcommon.a
+	$(top_builddir)/Common/libcommon.a -lz
 
 abyss_align_SOURCES = align.cc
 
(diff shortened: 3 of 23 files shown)


schutzekatze:
Right, the CI fails because of the zlib dependency which they don't include.

benvvalk:
@schutzekatze Thank you for sharing this!

Just to warn you, I think this work will require quite a bit of iteration and discussion before it can be included in the ABySS code base.  It is up to you if you want to follow through with that, of course!

Does `make` and `make check` run successfully on your own machine? As a bare minimum, you should ensure that those commands run to the end without errors.  Also, please fix any compiler warnings that are introduced by your changes.

I noticed that you commented out the `operator>>` method in `FastaReader.h` (?)  All of the ABySS programs use the `>>` operator as the main method to read in FASTA/FASTQ records (e.g. this line of the `abyss-map` program: https://github.com/bcgsc/abyss/blob/745e4a5aba0a273de2da7216a1b07a9fb29d5e6b/Map/map.cc#L486) .   In order to avoid changes to the approx 50 ABySS programs that use `FastaReader`, we would want  the `kseq` reading/parsing to be invoked through the `>>` operator, although I understand why incorporating that is difficult or perhaps not even possible in practice.  An alternative approach would be to profile/optimize/rewrite the existing parsing code in FastaReader.

Regarding CI:
* You can fix the zlib issue by adding zlib package(s) to the `apt-get` command in .travis.yml.  Eventually there should also be checks added to `configure.ac` (source file for the `configure` command) to ensure that the user has zlib before compiling.  It is amazing, but some systems do not have zlib installed by default.
* There are currently some Bloom filter related unit test failures when compiling under clang (with CircleCI, I believe), which are not your fault and which you should ignore.

benvvalk:
My mistake!  I see that the main `operator>>` method is still active, it is just the manipulator one that you commented out (does that really need to be removed?).  Okay, I get how this works now.

Making decisions based on filename extensions gives me a bit of an icky feeling, but I don't have necessarily have a good solution (yet).

schutzekatze:
I ran the whole pipeline and it works.

As for the manipulator operator, the thing is, it works with the stream input which is not used when kseq.h runs, as it uses file instead. Compiler doesn't complain that anything uses the manip operator, which is why it's commented out.

As for the name extension, I agree it's ehhh. Although it falls back to original solution and works normally if it the extension is not right.

This pull request is just an attempt to add kseq, I'm not sure if it's worth merging, because the speed gains are not thaat much.

benvvalk:
Fair enough about removing the manipulator method! Small request: Could you please delete the lines for the method instead of commenting them out?  (We try to avoid leaving commented code in our source files, because it eventually gets messy.)

Now that I am starting to understand your changes better, I feel that you did a nice job.  I will try to make some time to test it and improve it.

I am surprised that you didn't see a big performance improvement; perhaps it is because most of the ABySS programs are bottlenecked by computation rather than I/O.

benvvalk:
Btw, if you add `time=1` to your `abyss-pe` command, you can time the individual commands in the pipeline.

schutzekatze:
Yeah I think larger datasets will benefit more. The ones I tried out so far didn't take more than 2hrs. The thing I realized is that, even though kseq is way faster than original reader (up to 100% actually) when it comes to fasta/fastq files, it doesn't come close (only up to 10% improvement) when dealing with .gz files as most time is spent on gunzipping. I also profiled the code to make sure I made no mistake and the actual reading part of kseq is indeed what takes most of the time.

schutzekatze:
I'll close this pull request for now, because it requires some more work. In particular, it should be its own branch I need to set configure.ac up.

sjackman:
Thanks for this work, Vladimir! I had assumed that `kseq` would a make big difference to run time. If decompression is the bottle neck in reading though, perhaps not. I wonder how the speed of `abyss-tofastq` compares with the old `FastaReader` and the `kseq` implementation.

> Most of the time dealing with fasta/fastq files is spent unzipping them (if they're zipped), but because of zlib, kseq.h allows for extracting while reading, instead of extracting the whole file, which is slightly faster.

ABySS pipes the output of `gunzip -c` into ABySS, so the decompression is happening in a separate process and in parallel with FastaReader.

> Because kseq.h doesn't have peek(), there's a filename based check to see whether the file is fasta/fastq, otherwise it falls back to original implementation.

We could probably add a `peek()` function to `kseq`.

sjackman:
Thanks, Ben. I hadn't seen that PR just yet. I've commented over at #245.

sjackman:
Superseded by PR #245

Issue #224, opened 2018-05-07
abyss-2.1.0:  cannot get openmpi recognized by configure


mmokrejs:
Hi,
  I just realized that with abyss-2.1.0 I am not getting ABYSS-P binary. Would you please check the config.log and the other files. I see some gcc-7.3.0-related strict checks failing. I do have have this problem with abyss-2.0.2, though. Openmpi is 3.0.1 but I do not think this is relevant.

[build.log.txt](https://github.com/bcgsc/abyss/files/1979308/build.log.txt)
[config.log.txt](https://github.com/bcgsc/abyss/files/1979309/config.log.txt)
[config.status.txt](https://github.com/bcgsc/abyss/files/1979310/config.status.txt)

mmokrejs:
```
$ pkg-config --list-all | grep -i mpi

ompi-fort                           Open MPI - Portable high-performance MPI implementation
orte                                Open MPI Run-Time Environment (ORTE) - Open MPI's run-time environment functionality
ompi-cxx                            Open MPI - Portable high-performance MPI implementation
ompi-f77                            Open MPI - Portable high-performance MPI implementation
ompi-f90                            Open MPI - Portable high-performance MPI implementation
ompi-c                              Open MPI - Portable high-performance MPI implementation
ompi                                Open MPI - Portable high-performance MPI implementation
```

```
$ pkg-config ompi --print-variables
exec_prefix
includedir
libdir
pcfiledir
pkgincludedir
prefix
$ pkg-config ompi --cflags
-pthread
$ pkg-config ompi --libs
-L/apps/gentoo/usr/lib -Wl,-rpath -Wl,/apps/gentoo/usr/lib -Wl,-rpath -Wl,/apps/gentoo/usr/lib64 -Wl,--enable-new-dtags -lmpi
```

benvvalk:
Hi @mmokrejs,

I noticed in your `build.log.txt` you have `--enable-mpich` -- that is probably the source of your woes.  MPICH is an alternative/competitor to OpenMPI, which you probably do not have installed on your system.

```
[...]

Issue #213, opened 2018-04-05
abyss-sealer: improve --help text


mmokrejs:
Hi,
  the help text and Readme.md are quiet whether sealer is single-threaded or not. Therefore, I cannot evaluate whether the runtime options are applied per thread or not. Good that the runtime text (shown below) shows the *units* because the Readme.md speaks of some integer and I had no idea in which units. It would be great to bring the Readme.md in sync.

  -b, --bloom-size=N           size of Bloom filter (e.g. '40G'). Required
                               when not using pre-built Bloom filter(s)

  While at it, do `sed -e 's/abyss-sealer-b/abyss-sealer -b/g'`

mmokrejs:
Yeah, from 
```
  -s, --search-mem=N           mem limit for graph searches; multiply by the
                               number of threads (-j) to get the total mem used
                               for graph traversal [500M]
```
I see at least some parts are multithreaded, but that still does not tell me how to split my physical RAM between `-s` and `-b` values.

mmokrejs:
I started meanwhile
`abyss-sealer -j 104 -b 2100G -S tt_16D1C3L12__abyss_128-8.fa -k25 -k50 -k75 -k100 -k125 -k150 -k200 -k250 -L 500 -B nolimit -s 29000 -o $prefix *.fa` and seems the `-b` affects just the first k-mer splitting step while the `-s` did not kick in yet? So I may not be at the risk of running out-of-memory? I forgot to include the `M` in `-s 29000` for Megabytes.

2100G + 104*29000M = 2100G + 3106G while I have only 3099G physically available.

There is no other output available yet.

![sealer_memoryusage](https://user-images.githubusercontent.com/3297852/38378964-900c747c-38ff-11e8-90f6-2c40067079a2.png)
![sealer_cpu_usage](https://user-images.githubusercontent.com/3297852/38378970-934d99c2-38ff-11e8-96d7-7261be991576.png)

mmokrejs:
The job completed after 

```
resources_used.cput=881:41:17
resources_used.mem=2202647628kb
resources_used.ncpus=104
resources_used.walltime=21:41:21
```

[...]

PR #245, opened 2018-07-30
Add kseq lib

schutzekatze:
So, this pull request passes make check locally and ./configure checks for the presence of zlib.

I used a basic script to compare reading fastq files with abyss-tofastq which reads the file and outputs to sdtout:

```sh
#!/bin/bash

input=ERR2026298_1.fastq

echo "Input:"
du -sh $input

prefix=/scratch/work/project/bio/nikvl/
sufix=/bin/abyss-tofastq

echo
echo "With kseq:"
abyss=abyss-install
for in in $(seq 1 5); do
    (time ($prefix$abyss$sufix $input >/dev/null) 2>&1) | grep real
done

echo
echo "Without kseq:"
abyss=abyss-original-install
for in in $(seq 1 5); do
    (time ($prefix$abyss$sufix $input >/dev/null) 2>&1) | grep real
done
```

And the results for a 1.5GiB file are:
Input:
1.5G	ERR2026298_1.fastq

With kseq:
real	0m4.949s
real	0m4.929s
real	0m4.935s
real	0m4.927s
real	0m4.932s

Without kseq:
real	0m9.348s
real	0m9.417s
real	0m9.468s
real	0m9.432s
real	0m9.440s

I have also compared the output of abyss-tofastq with and without kseq with diff and they were equal, so the reading is correct. I have tried, however, doing diff on the *-8.fa files (from the whole pipeline) of abyss with and without kseq, and they seem to differ, but I assume that's because the order of resulting contigs is not deterministic. Do you know of a way to make sure the resulting output is actually the same?

Other than that, for an input of about 30GiBs, the job lasted 80 minutes and the speedup was always around several minutes for every test, so in the end, FastaReader is not much of a bottleneck, but for very large datasets, it can likely give a more noticeable improvement.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ jobs:
       - checkout
       - run: |
           ./autogen.sh
-          ./configure CC=clang CXX=clang++ --with-mpi=/usr/lib/openmpi
+          ./configure CC=clang CXX=clang++ CXXFLAGS=-std=c++11 --with-mpi=/usr/lib/openmpi
       - run: make -j
       - run: make -j check
       - run: make -j distcheck
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ jobs:
     steps:
       - run: |
           apt-get update -qq
-          apt-get install -qq autoconf automake clang g++ libboost-dev libgtest-dev libopenmpi-dev libsparsehash-dev make pandoc gdb valgrind
+          apt-get install -qq autoconf automake clang g++ libboost-dev libgtest-dev libopenmpi-dev libsparsehash-dev zlib1g-dev make pandoc gdb valgrind
       - checkout
       - run: |
           ./autogen.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,6 @@ before_install:
   - sudo apt-get install -qq libboost-dev libgtest-dev libopenmpi-dev libsparsehash-dev zlib1g-dev
 script:
   - ./autogen.sh
-  - ./configure --with-mpi=/usr/lib/openmpi
+  - ./configure CXXFLAGS=-std=c++11 --with-mpi=/usr/lib/openmpi
   - make
   - make check
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ compiler:
   - gcc
 before_install:
   - sudo apt-get update -qq
-  - sudo apt-get install -qq libboost-dev libgtest-dev libopenmpi-dev libsparsehash-dev
+  - sudo apt-get install -qq libboost-dev libgtest-dev libopenmpi-dev libsparsehash-dev zlib1g-dev
 script:
   - ./autogen.sh
   - ./configure --with-mpi=/usr/lib/openmpi
--- a/DataLayer/FastaReader.cpp
+++ b/DataLayer/FastaReader.cpp
@@ -10,6 +10,12 @@
 #include <sstream>
 #include <vector>
 
+extern "C" {
+#include "kseq.h"
+}
+#include <zlib.h>
+KSEQ_INIT(gzFile, gzread)
+
 using namespace std;
 
 namespace opt {
@@ -44,18 +50,72 @@ ostream& FastaReader::die()
 	return cerr << m_path << ':' << m_line << ": error: ";
 }
 
+static bool is_file_fasta(string path) {
+	return endsWith(path, ".fasta") ||
+		   endsWith(path, ".fasta.gz") ||
+		   endsWith(path, ".fa") ||
+		   endsWith(path, ".fa.gz");
+}
+
+static bool is_file_fastq(string path) {
+	return endsWith(path, ".fastq") ||
+		   endsWith(path, ".fastq.gz") ||
+		   endsWith(path, ".fq") ||
+		   endsWith(path, ".fq.gz");
+}
+
 FastaReader::FastaReader(const char* path, int flags, int len)
-	: m_path(path), m_fin(path),
-	m_in(strcmp(path, "-") == 0 ? cin : m_fin),
+	: m_path(path),
 	m_flags(flags), m_line(0), m_unchaste(0),
 	m_end(numeric_limits<streamsize>::max()),
 	m_maxLength(len)
 {
-	if (strcmp(path, "-") != 0)
-		assert_good(m_fin, path);
-	if (m_in.peek() == EOF)
-		cerr << m_path << ':' << m_line << ": warning: "
-			"file is empty\n";
+	if (is_file_fasta(path) || is_file_fastq(path)) {
+		if (is_file_fastq(path)) {
+			q = true;
+		}
+	    is_fasta = true;
+		m_file = gzopen(path, "r");
+		m_file_state = 0;
+		seq_data = kseq_init((gzFile)m_file);
+	} else {
+		m_fin = new ifstream(path);
+		m_in = (strcmp(path, "-") == 0 ? &cin : m_fin);
+
+		if (strcmp(path, "-") != 0)
+			assert_good(*m_fin, path);
+		if (m_in->peek() == EOF)
+			cerr << m_path << ':' << m_line << ": warning: "
+				"file is empty\n";
+	}
+}
+
+FastaReader::~FastaReader()
+{
+	if (is_fasta) {
+		kseq_destroy((kseq_t*)seq_data);
+		gzclose((gzFile)m_file);
+	} else {
+		if (!m_in->eof()) {
+			std::string line;
+			getline(line);
+			die() << "expected end-of-file near\n"
+				<< line << '\n';
+			exit(EXIT_FAILURE);
+		}
+
+		delete m_fin;
+	}
+}
+
+int gzpeek(gzFile f)
+{
+    int c;
+
+    c = gzgetc(f);
+    gzungetc(c, f);
+
+    return c;
 }
 
 /** Split the fasta file into nsections and seek to the start
@@ -67,31 +127,64 @@ void FastaReader::split(unsigned section, unsigned nsections)
 	assert(strcmp(m_path, "-") != 0);
 	if (nsections == 1)
 		return;
-	// Move the get pointer to the first entry in this section and
-	// update the m_end if there is more than one section.
-	m_in.seekg(0, ios::end);
-	streampos length = m_in.tellg();
-	assert(length > 0);
-	streampos start = length * (section - 1) / nsections;
-	streampos end = length * section / nsections;
-	assert(end > 0);
-	if (end < length) {
-		m_in.seekg(end);
-		if (m_in.peek() == '>')
-			end += 1;
+
+	if (is_fasta) {
+		// Move the get pointer to the first entry in this section and
+		// update the m_end if there is more than one section.
+		gzseek((gzFile)m_file, 0L, SEEK_END);
+		long length = gztell((gzFile)m_file);
+		assert(length > 0);
+		long start = length * (section - 1) / nsections;
+		long end = length * section / nsections;
+		assert(end > 0);
+		if (end < length) {
+			gzseek((gzFile)m_file, 0, end);
+			if (gzpeek((gzFile)m_file) == '>')
+				end += 1;
+		}
+		m_file_end = end;
+		gzseek((gzFile)m_file, 0, start);
+		if (start > 0) {
+			char c;
+			for (;;) {
+				c = gzgetc((gzFile)m_file);
+				if (c == EOF || c == '>' || c == '\n') {
+					break;
+				}
+			}
+			if (gzeof((gzFile)m_file))
+				cerr << m_path << ':' << section << ": warning: "
+					"there are no contigs in this section\n";
+			gzungetc('>', (gzFile)m_file);
+		}
+		assert(m_file_end > 0);
+	} else {
+		// Move the get pointer to the first entry in this section and
+		// update the m_end if there is more than one section.
+		m_in->seekg(0, ios::end);
+		streampos length = m_in->tellg();
+		assert(length > 0);
+		streampos start = length * (section - 1) / nsections;
+		streampos end = length * section / nsections;
+		assert(end > 0);
+		if (end < length) {
+			m_in->seekg(end);
+			if (m_in->peek() == '>')
+				end += 1;
+		}
+		m_end = end;
+		m_in->seekg(start);
+		if (start > 0) {
+			m_in->ignore(numeric_limits<streamsize>::max(), '\n');
+			m_in->ignore(numeric_limits<streamsize>::max(), '>');
+			if (m_in->peek() == EOF)
+				cerr << m_path << ':' << section << ": warning: "
+					"there are no contigs in this section\n";
+			m_in->putback('>');
+		}
+		assert(m_end > 0);
+		assert(m_in->good());
 	}
-	m_end = end;
-	m_in.seekg(start);
-	if (start > 0) {
-		m_in.ignore(numeric_limits<streamsize>::max(), '\n');
-		m_in.ignore(numeric_limits<streamsize>::max(), '>');
-		if (m_in.peek() == EOF)
-			cerr << m_path << ':' << section << ": warning: "
-				"there are no contigs in this section\n";
-		m_in.putback('>');
-	}
-	assert(m_end > 0);
-	assert(m_in.good());
 }
 
 /** Return whether this read passed the chastity filter. */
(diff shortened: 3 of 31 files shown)


benvvalk:
Thanks for this, @schutzekatze :-)

Rather than creating a new pull request, could you please update the branch for your original pull request with these improvements? (That is the usual workflow, I think.)

For `configure.ac`, you can do a more thorough check using the `AX_CHECK_ZLIB` macro: https://www.gnu.org/software/autoconf-archive/ax_check_zlib.html.  I'm not sure if it is built in to the standard autoconf library -- you may  have to download the `.m4` file and add it to the ABySS source tree.  In general, just checking for a header file (`zlib.h`) is not enough -- you should also check that you can compile and link against the library.  `AC_CHECK_LIB` is the generic way to do that, but if you can find a macro like `AX_CHECK_ZLIB` that someone has already written, that is even better.

I still have not tested your code, but the only remaining criticism I have about your PR is the filename-based behaviour. IMO, it would be okay to open and close the files twice -- once to check the format, once to read the data.  In the case of `-` filename (STDIN), where you can't reopen the stream, I think it would reasonable to just pass it through to the old parsing code.

Thanks for your excellent work!

schutzekatze:
Yeah, you're right, filename based check is far from perfect. I'll see to implement that and better configure check. Expect an update in the following days.

benvvalk:
@schutzekatze Thanks for your patience and persistence!

schutzekatze:
Dog bless AX_CHECK_ZLIB. It automatically adds the -lz flag so there's no need to edit all the Makefiles.

benvvalk:
@schutzekatze Very nice! I will test this and get back to you within the next couple of days.

Sorry about the failing CircleCI checks -- it is very annoying for me as well.

.travis.yml
```diff
@@ -3,9 +3,9 @@ compiler:
   - gcc
 before_install:
   - sudo apt-get update -qq
-  - sudo apt-get install -qq libboost-dev libgtest-dev libopenmpi-dev libsparsehash-dev
+  - sudo apt-get install -qq libboost-dev libgtest-dev libopenmpi-dev libsparsehash-dev zlib1g-dev
 script:
   - ./autogen.sh
-  - ./configure --with-mpi=/usr/lib/openmpi
+  - ./configure CXXFLAGS=-std=c++11 --with-mpi=/usr/lib/openmpi
```
sjackman: Please remove this change.

DataLayer/FastaReader.h
```diff
@@ -24,16 +25,7 @@ class FastaReader {
...
-		{
-			if (!m_in.eof()) {
-				std::string line;
-				getline(line);
-				die() << "expected end-of-file near\n"
-					<< line << '\n';
-				exit(EXIT_FAILURE);
-			}
-		}
+		~FastaReader();
```
sjackman: Please keep this function as inline.

DataLayer/FastaReader.h
```diff
@@ -76,32 +88,38 @@ class FastaReader {
...
 
 		std::ostream& die();
 		bool isChaste(const std::string& s, const std::string& line);
 		void checkSeqQual(const std::string& s, const std::string& q);
 
 		const char* m_path;
-		std::ifstream m_fin;
-		std::istream& m_in;
+		std::ifstream* m_fin;
+		std::istream* m_in;
```
sjackman: Please leave both of these as objects rather than pointers. It's okay if the file is never opened when using `kseq`.

DataLayer/FastaReader.h
```diff
@@ -76,32 +88,38 @@ class FastaReader {
...
+
+		bool is_fasta = false; // Determines what method of reading to use
+		bool q = false;
```
sjackman: Please use a longer variable name.
Please prefix all variable names with m_.
Please use camel case. `m_isFasta` rather than `m_is_fasta`.

Sealer/sealer.cc
```diff
@@ -438,10 +438,10 @@ string merge(const Graph& g,
 
 		case NO_PATH:
 			assert(paths.empty());
-			if (result.foundStartKmer && result.foundGoalKmer)
+			if (result.foundStartKmer && result.foundGoalKmer) {
 #pragma omp atomic
 				++g_count.noPath;
-			else {
+			} else {
```
sjackman: Please remove this unrelated change.
I'd accept this change in a separate PR.

Sealer/sealer.cc
```diff
@@ -451,12 +451,13 @@ string merge(const Graph& g,
...
 					result.readMismatches > params.maxReadMismatches) {
-				if (result.pathMismatches > params.maxPathMismatches)
+				if (result.pathMismatches > params.maxPathMismatches) {
 #pragma omp atomic
 					++g_count.tooManyMismatches;
-				else
+				} else {
 #pragma omp atomic
 					++g_count.tooManyReadMismatches;
+				}
```
sjackman: Please remove this change.

.circleci/config.yml
```diff
@@ -6,11 +6,11 @@ jobs:
     steps:
       - run: |
           apt-get update -qq
-          apt-get install -qq autoconf automake clang g++ libboost-dev libgtest-dev libopenmpi-dev libsparsehash-dev make pandoc gdb valgrind
+          apt-get install -qq autoconf automake clang g++ libboost-dev libgtest-dev libopenmpi-dev libsparsehash-dev zlib1g-dev make pandoc gdb valgrind
       - checkout
       - run: |
           ./autogen.sh
-          ./configure CC=clang CXX=clang++ --with-mpi=/usr/lib/openmpi
+          ./configure CC=clang CXX=clang++ CXXFLAGS=-std=c++11 --with-mpi=/usr/lib/openmpi
```
sjackman: Please remove this change.

DataLayer/FastaReader.cpp
```diff
@@ -67,31 +118,64 @@ void FastaReader::split(unsigned section, unsigned nsections)
...
+				if (c == EOF || c == '>' || c == '\n') {
+					break;
+				}
+			}
+			if (gzeof((gzFile)m_file))
+				cerr << m_path << ':' << section << ": warning: "
+					"there are no contigs in this section\n";
+			gzungetc('>', (gzFile)m_file);
+		}
+		assert(m_file_end > 0);
```
sjackman: Please factor this code out into its own function.

DataLayer/FastaReader.cpp
```diff
@@ -136,90 +220,26 @@ Sequence FastaReader::read(string& id, string& comment,
...
-				}
-				goto next_record;
-			}
-			if (id.size() > 2 && id.rbegin()[1] != '/') {
-				// Add the read number to the ID.
-				id += '/';
-				id += comment[0];
-			}
-		}
+	if (is_fasta) {
```
sjackman: Please factor this code out into its own function.

sjackman:
Thanks for this work, Vladimir! I've wondered for a while how using `kseq` would affect performance.

`abyss-tofastq` is about twice as fast. That's cool! What sort of difference does it make to mapping one lane of Illumina reads to a draft assembly (say the *-3.fa file) with `abyss-map`?

> I have tried, however, doing diff on the *-8.fa files (from the whole pipeline) of abyss with and without kseq, and they seem to differ, but I assume that's because the order of resulting contigs is not deterministic.

I believe the results of ABySS should be deterministic. With MPI, I believe it should be deterministic, but I'm not 100% convinced of that. For example, it's possible that the orientation of unitigs may possibly vary with MPI. Without MPI, I believe the results should be identical from run to run.

> Other than that, for an input of about 30GiBs, the job lasted 80 minutes and the speedup was always around several minutes for every test, so in the end, FastaReader is not much of a bottleneck, but for very large datasets, it can likely give a more noticeable improvement.

The performance is a modest few percentage points faster. Since the performance improvement is modest, I feel that the additional code complexity of the current PR doesn't justify this level of performance improvement. Perhaps we can reduce that complexity though with some changes to the PR.

-----

`Common/Uncompress.h` supports reading gzip (and other format) files by piping, so we don't need the `zlib` support of `kseq`. It should be possible to use `KSEQ_INIT` with `istream*` and a wrapper around `istream::read`? This has a few advantages:

1. no zlib
2. the code can continue to use `istream` which should reduce the amount of change a lot
3. you can use `peek`

benvvalk:
Thanks, @sjackman. 

Unfortunately, I agree about the tradeoff between code complexity vs. performance benefit probably not being worth it for this PR.  For example, it feels a bit wrong to keep and use the old FASTA/FASTQ parsing code alongside kseq, but only make use of it in the case of reading from STDIN. That is probably asking for trouble.  

It is a tricky problem to incorporate kseq while keeping the downstream usage of FastaReader the same and I feel that @schutzekatze has made an admirable attempt.  But you are probably right that we would need to find a better approach to incorporate kseq.

schutzekatze:
I feel like the whole kseq approach is not justified. The performance difference isn't that much and there's a whole layer of adapting existing C++ code to a C library. In my opinion, I think it's more worth spending the time I have left at this summer internship to try to get Sealer to run on multiple nodes with MPI which should directly influence its performance.
If anything, we know reading files isn't a bottleneck, it's more about the processing.

sjackman:
In that case let's retire this PR. Thanks for this work all the same! I've been curious for a long time how much of a bottle neck FastaReader is. It's very interesting to me that abyss-tofastq takes half the time. Thanks, again!

mmokrejs:
Hi everybody, thank you @sjackman  and @benvvalk for your comments. I discussed with Vladimir meanwhile other, less technical details of his work and I shamelessly left the code review itself upon you.
I specially agree with Shaun that Vladimir should have not sneaked into the PR the unrelated code changes (added missed braces, unnecessarily requiring c++11 compiler, etc.). These should indeed be in a separate pull request, if at all.

Regarding the kseq-driven changes. I asked meanwhile Vladimir to re-test his changes on the full fish genome we have, where unzipping+k-mer splitting took 60% of total runtime (in the hash-based running mode, no Bloom-filter-based approach). I think the buffered-approach of kseq is an improvement and is worth of some form of merging. We use LustreFS here and reading in short blocks like 64kB is way to ineffective.

@schutzekatze Please add Readme.md change to this PR to document the new zlib dependency provided you are not going to follow the approach from https://github.com/bcgsc/abyss/pull/245#issuecomment-411857108 right away (which sounds plausible).

@schutzekatze I suggest running `strace -v -f -s 1024 -x` or similar to inspect what are the read block sizes before and after your changes.

I still think the code changes are helpful if this PR could be further tuned. Other tools (like java-based bbmap are much faster handling FASTQ files and performing various k-mer based operations). abys/sealer code is not ideal and I would like to support Vladimir's attempts to make it faster. I think performance improvements shall be traded for eventual slight increase in code complexity. Vladimir unleashed portions of dead code elsewhere and this PR is an improvement. I am concerned about the buffer approach as it is more friendly to large file operations and not killing networked filesystem servers. That counts in my opinion too.

I mentioned to Vladimir that you already mentioned once to me in the past that you do not like MPI and you want to phase it out. However, I am optimistic and hope that he can help you with the existing MPI-related bug reports already opened and although not wished, provide MPI-improvements to sealer (actually, a hybrid approach). Indeed, as you know, here is a summer student at a supercomputer centre, he picked this project because it is a large-scale project, sealer is taking enormous runtime time on our servers and OpenMP-only parallelization is just not enough. Even on the largest NUMA server which consists of 14 NUMA nodes the file-based Io is peaking at only 1/4 of the speed we commonly achieve via MPI across infiniband (supposedly due to all interrupts going through CPU0, slowing the whole thing). So, 112 local CPUs and 3.2TB of RAM are just not the best solution.

Last but not least, this is a Summer of HPC project, it is intended to teach students performance tuning and paralelization on real problems. Three weeks are being left (not even counting in the final reports to be written) so practically he will continue his training exercise and keep improving the PR's to better fit your reviews. Please do not feel pushed to do anything aside from usual code reviews (as you have done so far). I believe Vladimir will update the pull request branches (and open new PRs for those unrelated code fixes) and will base his final reports on those updated versions. It is somewhat expected eventual code merges shall happen after the SoHPC is over or never at all if they do not pass the reviews. This is a training opportunity and it is all expected.

Incidentally I also discussed with Vladimir some suposedly graph-related perfomance issues but it felt like too much work for the remaining time. However, I suggested to visualize the networked graph and show median values of alternatives paths through the graph so one could better or even automatically tune the `--max-paths=2` threshold for merging of alternative alleles into a single one (https://github.com/bcgsc/abyss/issues/213).

sjackman:
> I think the buffered-approach of kseq is an improvement and is worth of some form of merging. We use LustreFS here and reading in short blocks like 64kB is way too ineffective.

`istream` does indeed use buffers when reading. The size of the buffer can even be changed using `rdbuf()->pubsetbuf(buffer, size)`. I'm not sure what the default buffer size for GCC, but I'd be curious to know if you figure it out.

Here's an interesting StackOverflow question on attempting to get iostream to perform better.
See https://stackoverflow.com/questions/5166263/how-to-get-iostream-to-perform-better

sjackman:
> However, I am optimistic and hope that he can help you with the existing MPI-related bug reports already opened

This issue https://github.com/bcgsc/abyss/issues/224 should be pretty straight forward to fix, and we'd welcome a PR to fix it.

sjackman:
> sealer is taking enormous runtime time on our servers and OpenMP-only parallelization is just not enough.

Do you think the construction of the Bloom filter or sealing the gaps (navigating the dBG) is the bottle neck? How long do these two stages take for your data set?

mmokrejs:
>> sealer is taking enormous runtime time on our servers and OpenMP-only parallelization is just not enough.

> Do you think the construction of the Bloom filter or sealing the gaps (navigating the dBG) is the bottle neck? How long do these two stages take for your data set?

I do not remember the detais but Vladimir will have the details after rerunning sealer soon.

I would appreciate more detailed logging (timestamps on the lines logged to STDOUT) and more often some kind of a snapshotting or output of intermediate results (ideally after every k-mer value completed). I think sealer does not output any temp results after completing a single k-mer value. It finished successfully even with `abyss-sealer -j 104 -b 2100G -k25 -k50 -k75 -k100 -k125 -k150 -k200 -k250 -L 500 -B nolimit -s 29000 ...`. Per logs I ended up with 10.8873% gaps closed. Admittedly the `-B nolimit` is a bit of an overkill but I wanted to get the best possible result out.

sjackman:
Yes, intermediate results could be helpful in troubleshooting. We can't write out the contigs as they're sealed, because it may have one gap closed in one value of *k*, and another gap closed in a different value of *k*. We could however write out the sequence of each closed gap to a different file. That'd be a useful feature, and we'd be happy to accept a PR to implement it.

Writing timestamps at the start of each line of output would be universally useful throughout the ABySS pipeline. We likely won't get to it ourselves, but you can pipe the output of ABySS to the `ts` (timestamp) utility of `moreutils`. I recommend the one true date format, ISO 8601.
https://unix.stackexchange.com/questions/26728/prepending-a-timestamp-to-each-line-of-output-from-a-command

ISO 8601
https://xkcd.com/1179/

schutzekatze:
Writing sequences of closed gaps sounds fairly straightforward, so I'll attempt to implement that.
I assume they'd all be written to a single file, each with FASTA id of the gap that should be filled, right?

Also, I ended up writing an MPI version of Sealer that scales with both k values and gaps as practice, as this summer school is centered around learning how to run jobs on multiple nodes. I understand that you're removing MPI support, so I won't be opening a pull request for that. Martin Mokrejs will probably have use for it, however, until you're done with the transition to abyss-bloom-dbh.

Cheers.

PS: xkcd rules

benvvalk:
@schutzekatze Sounds good! There may be multiple gaps per input FASTA sequence, so you would probably want to suffix the FASTA ID with the gap start position, in order to uniquely identify each gap-filling sequence.

schutzekatze:
Makes sense. Do you think I should add a command line option for the closed gap filename and only output it if the option is present?

benvvalk:
@schutzekatze Yep, that's our convention in ABySS -- each option for generating an auxiliary file requires a filename, and an auxiliary file is only generated if its option/filename is explicitly specified by the user.  It is a good convention in general, because it prevents cluttering up the user's directory with mystery files.

mmokrejs:
I would prefer one FASTA file per k-mer value, instead of having zillions of tiny files. Ideally also with relative size as percentages of the gap actually closed.

>$original_scaffold_or_contig_id:$start-$stop $relative_size_of_the_gap_closed
atcg

@schutzekatze Please do make a pull request for the MPI support, it is only good if you get a review of your changes, other people may join in later, and it is a clear documentation of your work.

@sjackman Thank you for mentioning `ts` utility, I did not know about it.Could it be incorporated into the wrapper Makefile?

sjackman:
I'd suggest piping the output of `abyss-pe` into `ts`. For example
```sh
nohup sh -c 'time abyss-pe … 2>&1 | ts' >abyss.log
```

sjackman:
> I assume they'd all be written to a single file, each with FASTA id of the gap that should be filled, right?

I belive Vladimir is suggesting one file in total.

For the FASTA ID I'd suggest…

```
${original_scaffold_or_contig_id}_${start}-${end} LN:i:…
```
`${end}-${start}+1` (1 based inclusive) gives the original length of the gap. `LN:i:…` gives the new length of the gap.

sjackman:
We likely won't merge the MPI PR, but you're welcome to open it, to keep it on record. Congrats on completing the work!

mmokrejs:
> `${original_scaffold_or_contig_id}_${start}-${end}`

I assume some people may input the `${original_scaffold_or_contig_id}` contaning underscores already so I would still prefer `:` as the delimiter from `${start}-${end}` but it is a matter of preference. 

Good that there will be oly a handful new FASTA fies, I indeed misread the original message.

sjackman:
I prefer not to use `:` in sequence IDs, because `:` is used by `samtools view` and IGV as the delimiter separating the sequence ID from the coordinate range. If the contig ID itself has a colon in it, things go poorly.

schutzekatze:
Can you just clarify the "LN:i:..."? What's i here and where does the new gap length go, exactly?

mmokrejs:
> I prefer not to use : in sequence ID ...

OK, agreed.

sjackman:
`LN:i:` is the SAM tag for sequence length. `i` is integer type.
See https://github.com/samtools/hts-specs
and https://github.com/samtools/hts-specs/blob/master/SAMtags.pdf

For example:
```
>1_1001-1100 LN:i:8
ACGTACGT
```
The gap of 100 Ns at `1:1001-1100` is replaced by the 8 characters `ACGTACGT`.

PR #247, opened 2018-08-16
Add closed gap file output

schutzekatze:
It's fairly straightforward and after running a job, the output looks good, the gaps are there with proper IDs.

### DIFF
--- a/Sealer/sealer.cc
+++ b/Sealer/sealer.cc
@@ -110,8 +110,8 @@ static const char USAGE_MESSAGE[] =
 "  -s, --search-mem=N           mem limit for graph searches; multiply by the\n"
 "                               number of threads (-j) to get the total mem used\n"
 "                               for graph traversal [500M]\n"
-"  -t, --trace-file=FILE        write graph search stats to FILE\n"
 "  -g, --gap-file=FILE          write sealed gaps to FILE\n"
+"  -t, --trace-file=FILE        write graph search stats to FILE\n"
 "  -v, --verbose                display verbose output\n"
 "      --help                   display this help and exit\n"
 "      --version                output version information and exit\n"
@@ -240,7 +240,7 @@ struct Counters {
 	size_t skipped;
 };
 
-static const char shortopts[] = "S:L:b:B:d:ef:F:G:i:Ij:k:lm:M:no:P:q:r:s:t:g:v";
+static const char shortopts[] = "S:L:b:B:d:ef:F:G:g:i:Ij:k:lm:M:no:P:q:r:s:t:v";
 
 enum { OPT_HELP = 1, OPT_VERSION };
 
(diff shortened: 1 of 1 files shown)


Sealer/sealer.cc
```diff
@@ -615,6 +621,13 @@ void kRun(const ConnectPairsParams& params,
 				uniqueGapsClosed++;
 				if (gapsclosed % 100 == 0)
 					printLog(logStream, IntToString(gapsclosed) + " gaps closed so far\n");
+
+				if (!opt::gapfilePath.empty()) {
+					gapStream << ">" << read1.id.substr(0,read1.id.length()-2)
+						  << "_" << read2_it->second.gapStart() << "-" << read2_it->second.gapEnd()
+						  << " LN:i:" << tempSeq.length() << endl;
+					gapStream << tempSeq << endl;
```
sjackman: Change these two lines to…
```c++
<< " LN:i:" << tempSeq.length() << '\n'
<< gapStream << tempSeq << '\n';
```
`std::endl` flushes the stream's buffer, which is undesirable.

Sealer/sealer.cc
```diff
@@ -874,6 +889,16 @@ int main(int argc, char** argv)
 		assert_good(traceStream, opt::tracefilePath);
 	}
 
+	ofstream gapStream;
+	if (!opt::gapfilePath.empty()) {
+		if (opt::verbose)
+			cerr << "Writing sealed gaps to '"
+				<< opt::gapfilePath << "'\n";
+		gapStream.open(opt::gapfilePath.c_str());
+		assert(gapStream.is_open());
```
sjackman: Please remove this line.

sjackman:
So as far as I can tell, the function that seals the gaps `kRun` is not parallelized at all! @benvvalk True?

@schutzekatze Would you be interested in parallelizing `kRun` using OpenMP?

schutzekatze:
I have just realized there are two connectPairs functions in Konnector (one in konnector.h, one in konnector.cc), used by Sealer to seal gaps, one of which is multithreaded and one of which is not, and Sealer is using the non-multithreaded one. I'll see if I have enough time to add omp parallel to it.

Sealer/sealer.cc
schutzekatze: I realize that it does, I added endl there because this file is supposed to give intermediate results, meaning they should be flushed to the file as soon as the gap is closed so that you can inspect it. But sure, I'll change it.

benvvalk:
Merged. Thank you, @schutzekatze!

Thanks for doing the code review, @sjackman!

benvvalk:
@schutzekatze I see what you mean about the multithreading (or lack thereof).  Good eye!

Please start a separate pull request for that fix (and thank you!)

sjackman:
While you're in there, limiting the maximum number of vertices visited per gap search would help a lot in reducing the worst-case run time of the algorithm. With those two changes (threading and cost limit), it's possible that we could enable Sealer by default in abyss-pe.

PR #248, opened 2018-08-18
Add MPI functionality to Sealer

schutzekatze:
Here's the MPI version of Sealer I was working on, in case it ends up being useful. (I understand MPI is being phased out, I am posting this in case somebody has super long Sealer jobs until then, and it was good practice for me.)

The way it works is it splits the processes into groups, with each group closing gaps for a particular value of k, so it scales by both the number of k values and the number of gaps to be closed. After k values are processed by all groups, closed gaps are compared and corrections are made so that the gaps closed by later values of k are erased if the gaps were closed by earlier k values. After all k values are processed, the closed gaps are gathered at root process which then writes it to a file, as before. Processes also each have their own log files which are merged in the end into a single file.

For example, for a Sealer job that lasts 16 minutes, time was brought down to about 3 minutes for a 7 node job, where each k value gets its own node (by default, there are 7 k values for Sealer).

I have also tested various number of nodes for the job, ranging from 1 to 20 and the output is identical to the output of original Sealer.

### DIFF
--- a/Sealer/Makefile.am
+++ b/Sealer/Makefile.am
@@ -15,7 +15,8 @@ abyss_sealer_CXXFLAGS = $(AM_CXXFLAGS) $(OPENMP_CXXFLAGS)
 abyss_sealer_LDADD = \
 	$(top_builddir)/DataLayer/libdatalayer.a \
 	$(top_builddir)/Align/libalign.a \
-	$(top_builddir)/Common/libcommon.a
+	$(top_builddir)/Common/libcommon.a \
+	$(MPI_LIBS)
 
 abyss_sealer_SOURCES = sealer.cc \
 	$(top_srcdir)/Bloom/BloomFilter.h \
--- a/Sealer/sealer.cc
+++ b/Sealer/sealer.cc
@@ -881,7 +881,7 @@ void send_string(const string& str, int target_pid) {
 	int chars_len = str.length() + 1; // +1 for NULL terminator
 	MPI_Send(&chars_len, 1, MPI_INT, target_pid, 0, MPI_COMM_WORLD);
 	const char *chars = str.c_str();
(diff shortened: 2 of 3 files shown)


benvvalk:
@schutzekatze Thanks for sharing your work!

This code is quite scary and complex compared to the (hypothetical) equivalent implementation with OpenMP!

schutzekatze:
Indeed!

I was looking into OpenMP parallelization of Sealer and it requires some knowledge of how gap sealing works in order to do it correctly, so I am not yet sure how simple it is. I also expected this MPI version to be way simpler when I started doing it, but ended up being somewhat complex, in order for it to work correctly. In any case, it's just a possible implementation of parallelization.

What I have learned, though, is that _a lot_ of the time is spent building the Bloom filters, and parallelizing by k values helps a lot, as filters for each k value is built concurrently as opposed to processing values 1 by 1.

I am not sure I have sufficient time left for the OpenMP parallelization, though, as the rest of this summer school is reserved for writing a report for what's been done.

sjackman:
If you do end up looking into parallelizing the gap sealing (graph navigation) step of Sealer, start with…
```diff
+#pragma omp parallel
	for (read1_it = flanks.begin(); read1_it != flanks.end();) {
```
https://github.com/bcgsc/abyss/blob/9aceb97de2925b20e30786999fe978a48a8b9f8d/Sealer/sealer.cc#L606
and then add appropriate locks `#pragma omp critical(name-of-lock)` around the shared output variables and streams.

schutzekatze:
In addition to opening the other pull request for adding OpenMP to Sealer, I have separately added OpenMP to this branch, because the addition slightly differs.

sjackman:
Supersed by PR https://github.com/bcgsc/abyss/pull/249 `Add OpenMP support to gap sealing`

PR #249, opened 2018-08-20
Add OpenMP support to gap sealing

schutzekatze:
This pull request parallelizes gap sealing in Sealer with OpenMP by giving each thread its own group of gaps to process from the total pool.

I am doing extra tests, but on a small dataset the output is identical. The speedup depends on how much time is spent on building bloom filter compared to actually sealing gaps.
I added two const keywords to the merge function to make sure it's threadsafe, apart from modifying g_count and traceStream, both of which are protected by omp critical and atomic.

### DIFF
--- a/Sealer/sealer.cc
+++ b/Sealer/sealer.cc
@@ -420,8 +420,8 @@ template <typename Graph>
 string merge(const Graph& g,
 	unsigned k,
 	const Gap& gap,
-	FastaRecord &read1,
-	FastaRecord &read2,
+	const FastaRecord &read1,
+	const FastaRecord &read2,
 	const ConnectPairsParams& params,
 	Counters& g_count,
 	ofstream& traceStream)
@@ -583,7 +583,6 @@ void kRun(const ConnectPairsParams& params,
 	map<FastaRecord, map<FastaRecord, Gap> >::iterator read1_it;
 	map<FastaRecord, Gap>::iterator read2_it;
 	unsigned uniqueGapsClosed = 0;
-	bool success;
 
 	Counters g_count;
 	g_count.noStartOrGoalKmer = 0;
@@ -603,38 +602,50 @@ void kRun(const ConnectPairsParams& params,
 
 	printLog(logStream, "Flanks inserted into k run = " + IntToString(flanks.size()) + "\n");
 
-	for (read1_it = flanks.begin(); read1_it != flanks.end();) {
-		success = false;
+	int counter = 0;
+	vector<map<FastaRecord, map<FastaRecord, Gap> >::iterator> flanks_closed;
+#pragma omp parallel private(read1_it, read2_it) firstprivate(counter)
+	for (read1_it = flanks.begin(); read1_it != flanks.end(); ++read1_it) {
 		FastaRecord read1 = read1_it->first;
-		for (read2_it = flanks[read1].begin(); read2_it != flanks[read1].end(); read2_it++) {
+		bool success = false;
+		for (read2_it = flanks[read1].begin(); read2_it != flanks[read1].end(); ++read2_it, ++counter) {
+#if _OPENMP
+			if (counter % omp_get_num_threads() != omp_get_thread_num())
+				continue;
+#endif
 			FastaRecord read2 = read2_it->first;
 
 			int startposition = read2_it->second.gapStart();
 			string tempSeq = merge(g, k, read2_it->second, read1, read2, params, g_count, traceStream);
 			if (!tempSeq.empty()) {
 				success = true;
+#pragma omp critical (allmerged)
 				allmerged[read1.id.substr(0,read1.id.length()-2)][startposition]
 					= ClosedGap(read2_it->second, tempSeq);
-//#pragma omp atomic
-				gapsclosed++;
-//#pragma omp atomic
-				uniqueGapsClosed++;
-				if (gapsclosed % 100 == 0)
+#pragma omp atomic
+				++uniqueGapsClosed;
+#pragma omp critical (gapsclosed)
+				if (++gapsclosed % 100 == 0)
 					printLog(logStream, IntToString(gapsclosed) + " gaps closed so far\n");
 
-				if (!opt::gapfilePath.empty()) {
+				if (!opt::gapfilePath.empty())
+#pragma omp critical (gapStream)
 					gapStream << ">" << read1.id.substr(0,read1.id.length()-2)
 						  << "_" << read2_it->second.gapStart() << "-" << read2_it->second.gapEnd()
-						  << " LN:i:" << tempSeq.length() << '\n';
-					gapStream << tempSeq << '\n';
-				}
+						  << " LN:i:" << tempSeq.length() << '\n'
+						  << tempSeq << '\n';
 			}
 		}
 		if (success) {
-			flanks.erase(read1_it++);
+#pragma omp critical (flanks_closed)
+			flanks_closed.push_back(read1_it);
 		}
-		else
-			read1_it++;
+	}
+
+	for (vector<map<FastaRecord, map<FastaRecord, Gap> >::iterator>::iterator it = flanks_closed.begin();
+			it != flanks_closed.end(); ++it) {
+		if (flanks.count((*it)->first) > 0)
+		    flanks.erase(*it);
 	}
 
 	printLog(logStream, IntToString(uniqueGapsClosed) + " unique gaps closed for k" + IntToString(k) + "\n");


benvvalk:
At first glance, this looks good. Thanks for this, @schutzekatze!  Looking forward to trying this.

Small request: Can you please line up your `#pragma`s with column 0 (i.e. remove identation for `#pragma` lines)? That is our convention in the rest of the ABySS code.

Sealer/sealer.cc
```diff
@@ -603,38 +602,57 @@ void kRun(const ConnectPairsParams& params,
...
 	printLog(logStream, "Flanks inserted into k run = " + IntToString(flanks.size()) + "\n");
 
-	for (read1_it = flanks.begin(); read1_it != flanks.end();) {
-		success = false;
+	int counter = 0;
+	vector<map<FastaRecord, map<FastaRecord, Gap> >::iterator> flanks_closed;
+#pragma omp parallel private(read1_it, read2_it) firstprivate(counter)
+	for (read1_it = flanks.begin(); read1_it != flanks.end(); read1_it++) {
 		FastaRecord read1 = read1_it->first;
+		bool success = true;
```
sjackman: Should this line be `bool success = false;`?

Sealer/sealer.cc
```diff
@@ -603,38 +602,57 @@ void kRun(const ConnectPairsParams& params,
 
 	printLog(logStream, "Flanks inserted into k run = " + IntToString(flanks.size()) + "\n");
 
-	for (read1_it = flanks.begin(); read1_it != flanks.end();) {
-		success = false;
+	int counter = 0;
+	vector<map<FastaRecord, map<FastaRecord, Gap> >::iterator> flanks_closed;
+#pragma omp parallel private(read1_it, read2_it) firstprivate(counter)
```
sjackman: I sit possible to use `#pragma omp parallel for` rather than `counter % omp_get_num_threads() == omp_get_thread_num()`? Is it possible to parallelize a for loop that's iterating over a STL data structure like `std::map`?

Sealer/sealer.cc
```diff
@@ -603,38 +602,57 @@ void kRun(const ConnectPairsParams& params,
...
-				if (gapsclosed % 100 == 0)
-					printLog(logStream, IntToString(gapsclosed) + " gaps closed so far\n");
-
-				if (!opt::gapfilePath.empty()) {
-					gapStream << ">" << read1.id.substr(0,read1.id.length()-2)
-						  << "_" << read2_it->second.gapStart() << "-" << read2_it->second.gapEnd()
-						  << " LN:i:" << tempSeq.length() << '\n';
-					gapStream << tempSeq << '\n';
+			#ifdef _OPENMP
+			if (counter % omp_get_num_threads() == omp_get_thread_num()) {
```
sjackman: To avoid adding another level of indentation (and this function already has too many) I'd suggest instead…
```c++
if (counter % omp_get_num_threads() != omp_get_thread_num())
    continue;
```

Sealer/sealer.cc
```diff
@@ -603,38 +602,57 @@ void kRun(const ConnectPairsParams& params,
...
-				uniqueGapsClosed++;
-				if (gapsclosed % 100 == 0)
-					printLog(logStream, IntToString(gapsclosed) + " gaps closed so far\n");
-
-				if (!opt::gapfilePath.empty()) {
-					gapStream << ">" << read1.id.substr(0,read1.id.length()-2)
-						  << "_" << read2_it->second.gapStart() << "-" << read2_it->second.gapEnd()
-						  << " LN:i:" << tempSeq.length() << '\n';
-					gapStream << tempSeq << '\n';
+			#ifdef _OPENMP
```
sjackman: Use `#if _OPENMP`

Sealer/sealer.cc
```diff
@@ -603,38 +602,57 @@ void kRun(const ConnectPairsParams& params,
...
+					allmerged[read1.id.substr(0,read1.id.length()-2)][startposition]
+						= ClosedGap(read2_it->second, tempSeq);
+#pragma omp atomic
+					uniqueGapsClosed++;
+#pragma omp critical (gapsclosed)
+					{
+						gapsclosed++;
+						if (gapsclosed % 100 == 0)
+							printLog(logStream, IntToString(gapsclosed) + " gaps closed so far\n");
+					}
```
sjackman: I'd suggest…
```c++
#pragma omp critical (gapsclosed)
						if (++gapsclosed % 100 == 0)
 							printLog(logStream, IntToString(gapsclosed) + " gaps closed so far\n");
```

Sealer/sealer.cc
```diff
@@ -603,38 +602,57 @@ void kRun(const ConnectPairsParams& params,
...
+
+					if (!opt::gapfilePath.empty())
+#pragma omp critical (gapfilePath)
+					{
+						gapStream << ">" << read1.id.substr(0,read1.id.length()-2)
+							  << "_" << read2_it->second.gapStart() << "-" << read2_it->second.gapEnd()
+							  << " LN:i:" << tempSeq.length() << '\n';
+						gapStream << tempSeq << '\n';
+					}
```
sjackman: Please remove the block and make this a single statement.

sjackman:
This work looks good! Thanks, Valdimir!

Sealer/sealer.cc
schutzekatze: I'll see if this works, actually. I thought it was not supported, but reading up on it, it seems it should be.

schutzekatze: It seems comparing map iterators with less than operator (<) is not valid (for the loop exit condition), but it is required for a loop to be parallelized with OpenMP, as it needs a "countable" loop. Map iterators seem to only allow inequality (!=) comparison, which is the case right now.

sjackman: How about storing the keys of the map in a vector, and then `parallel omp for` iterate over that vector?

Sealer/sealer.cc
```diff
@@ -603,38 +602,48 @@ void kRun(const ConnectPairsParams& params,
 
 	printLog(logStream, "Flanks inserted into k run = " + IntToString(flanks.size()) + "\n");
 
-	for (read1_it = flanks.begin(); read1_it != flanks.end();) {
-		success = false;
+	int counter = 0;
+	vector<map<FastaRecord, map<FastaRecord, Gap> >::iterator> flanks_closed;
+#pragma omp parallel private(read1_it, read2_it) firstprivate(counter)
+	for (read1_it = flanks.begin(); read1_it != flanks.end(); read1_it++) {
```
sjackman: Please use preincrement `++read1_it` for style (and theoretical performance benefit, though probably not actual).

Sealer/sealer.cc
```diff
@@ -603,38 +602,48 @@ void kRun(const ConnectPairsParams& params,
...
 		FastaRecord read1 = read1_it->first;
-		for (read2_it = flanks[read1].begin(); read2_it != flanks[read1].end(); read2_it++) {
+		bool success = false;
+		for (read2_it = flanks[read1].begin(); read2_it != flanks[read1].end(); read2_it++, counter++) {
```
sjackman: Use preincrement operator please.

Sealer/sealer.cc
```diff
@@ -603,38 +602,48 @@ void kRun(const ConnectPairsParams& params,
...
+			if (counter % omp_get_num_threads() == omp_get_thread_num())
+				continue;
+			#endif
 			FastaRecord read2 = read2_it->first;
 
 			int startposition = read2_it->second.gapStart();
 			string tempSeq = merge(g, k, read2_it->second, read1, read2, params, g_count, traceStream);
 			if (!tempSeq.empty()) {
 				success = true;
+#pragma omp critical (allmerged)
```
sjackman: If this lock were a bottleneck, each thread could have its own private copy of `allmerged`, which were the merged at the end. Please don't make this change however. I prefer the current code for now. Just an observation.

Sealer/sealer.cc
```diff
@@ -603,38 +602,48 @@ void kRun(const ConnectPairsParams& params,
...
 				allmerged[read1.id.substr(0,read1.id.length()-2)][startposition]
 					= ClosedGap(read2_it->second, tempSeq);
-//#pragma omp atomic
-				gapsclosed++;
-//#pragma omp atomic
+#pragma omp atomic
 				uniqueGapsClosed++;
```
sjackman: Preincrement operator please.

Sealer/sealer.cc
```diff
@@ -603,38 +602,48 @@ void kRun(const ConnectPairsParams& params,
...
 
-				if (!opt::gapfilePath.empty()) {
+				if (!opt::gapfilePath.empty())
+#pragma omp critical (gapfilePath)
 					gapStream << ">" << read1.id.substr(0,read1.id.length()-2)
 						  << "_" << read2_it->second.gapStart() << "-" << read2_it->second.gapEnd()
-						  << " LN:i:" << tempSeq.length() << '\n';
-					gapStream << tempSeq << '\n';
-				}
+						  << " LN:i:" << tempSeq.length() << '\n' << tempSeq << '\n';
```
sjackman: Please put `<< tempSeq << '\n';` on its own line.

Sealer/sealer.cc
```diff
@@ -603,38 +602,48 @@ void kRun(const ConnectPairsParams& params,
...
+#pragma omp critical (flanks_closed)
+			flanks_closed.push_back(read1_it);
 		}
-		else
-			read1_it++;
+	}
+
+	for (vector<map<FastaRecord, map<FastaRecord, Gap> >::iterator>::iterator it = flanks_closed.begin();
+			it != flanks_closed.end(); it++) {
+		flanks.erase(*it);
```
sjackman: I'm concerned that this code may not be valid semantically. It's a sticky part of C++, but I don't believe that you're permitted to `erase` an iterator that your actively using for iteration. It's different for each data structure though. you may need something like
```c++
for (vector<map<FastaRecord, map<FastaRecord, Gap> >::iterator>::iterator
 			it = flanks_closed.begin();
 			it != flanks_closed.end();)
 	it = flanks.erase(it);
```
You'll find posts on the topic when you Google for it.

sjackman: What is the purpose though of erasing these elements from `flank`?

Sealer/sealer.cc
```diff
@@ -603,38 +602,48 @@ void kRun(const ConnectPairsParams& params,
...
-		success = false;
+	int counter = 0;
+	vector<map<FastaRecord, map<FastaRecord, Gap> >::iterator> flanks_closed;
+#pragma omp parallel private(read1_it, read2_it) firstprivate(counter)
+	for (read1_it = flanks.begin(); read1_it != flanks.end(); read1_it++) {
 		FastaRecord read1 = read1_it->first;
-		for (read2_it = flanks[read1].begin(); read2_it != flanks[read1].end(); read2_it++) {
+		bool success = false;
+		for (read2_it = flanks[read1].begin(); read2_it != flanks[read1].end(); read2_it++, counter++) {
+			#if _OPENMP
```
sjackman: Please undent all `#` directives.

Sealer/sealer.cc
```diff
@@ -603,38 +602,48 @@ void kRun(const ConnectPairsParams& params,
...
+			if (counter % omp_get_num_threads() == omp_get_thread_num())
+				continue;
+			#endif
```
sjackman: Please undent all `#` directives.

Sealer/sealer.cc
```diff
@@ -603,38 +602,48 @@ void kRun(const ConnectPairsParams& params,
...
+	int counter = 0;
+	vector<map<FastaRecord, map<FastaRecord, Gap> >::iterator> flanks_closed;
+#pragma omp parallel private(read1_it, read2_it) firstprivate(counter)
+	for (read1_it = flanks.begin(); read1_it != flanks.end(); read1_it++) {
 		FastaRecord read1 = read1_it->first;
-		for (read2_it = flanks[read1].begin(); read2_it != flanks[read1].end(); read2_it++) {
+		bool success = false;
+		for (read2_it = flanks[read1].begin(); read2_it != flanks[read1].end(); read2_it++, counter++) {
+			#if _OPENMP
+			if (counter % omp_get_num_threads() == omp_get_thread_num())
```
sjackman: Shouldn't this be…
```c++
if (counter % omp_get_num_threads() != omp_get_thread_num())
```
?

Sealer/sealer.cc
```diff
@@ -603,38 +602,48 @@ void kRun(const ConnectPairsParams& params,
...
+#pragma omp atomic
 				uniqueGapsClosed++;
-				if (gapsclosed % 100 == 0)
+#pragma omp critical (gapsclosed)
+				if (++gapsclosed % 100 == 0)
 					printLog(logStream, IntToString(gapsclosed) + " gaps closed so far\n");
 
-				if (!opt::gapfilePath.empty()) {
+				if (!opt::gapfilePath.empty())
+#pragma omp critical (gapfilePath)
```
sjackman: Please change this line to
```c++
#pragma omp critical (gapStream)
```

Sealer/sealer.cc
```diff
@@ -603,38 +602,48 @@ void kRun(const ConnectPairsParams& params,
...
-						  << " LN:i:" << tempSeq.length() << '\n';
-					gapStream << tempSeq << '\n';
-				}
+						  << " LN:i:" << tempSeq.length() << '\n' << tempSeq << '\n';
 			}
 		}
 		if (success) {
-			flanks.erase(read1_it++);
+#pragma omp critical (flanks_closed)
+			flanks_closed.push_back(read1_it);
```
sjackman: Similar comment to above, each thread could have a private copy of `flanks_closed` to avoid lock contention, but please don't change it now.

Sealer/sealer.cc
```diff
@@ -603,38 +602,48 @@ void kRun(const ConnectPairsParams& params,
...
+#pragma omp critical (allmerged)
 				allmerged[read1.id.substr(0,read1.id.length()-2)][startposition]
 					= ClosedGap(read2_it->second, tempSeq);
-//#pragma omp atomic
-				gapsclosed++;
-//#pragma omp atomic
+#pragma omp atomic
 				uniqueGapsClosed++;
-				if (gapsclosed % 100 == 0)
+#pragma omp critical (gapsclosed)
```
sjackman: I'm inclined to change this line to…
```c++
#pragma omp critical (logStream)
```
Either would be correct currently, but just in case someone adds more printing to `logStream` in the future.

Sealer/sealer.cc
schutzekatze: Yep, forgot to invert the operator after switching to continue.

Sealer/sealer.cc
schutzekatze: For storing keys in a vector, it seems it would be possible to fully replace the map there with a vector in the first place, while building flanks (Actually, is that what you meant?). Otherwise, it might be a bit slow to transfer keys to a vector.

Sealer/sealer.cc
schutzekatze: Don't think it's a bottleneck, though, most of the work is done in the merge function.

Sealer/sealer.cc
schutzekatze: It seems it would be possible to fully replace the map there with a vector in the first place, while building flanks (Actually, is that what you meant?). Otherwise, it might be a bit slow to transfer keys to a vector.

sjackman: Well, that's the better ultimate solution, but it'd make for a bigger diff that's more difficult to review. I meant copy the keys to a second `std:vector` and iterate over that. Let's stick with what you have, using the `counter % omp_get_num_threads() == omp_get_thread_num()`, for this PR. After we merge this PR, I'd appreciate as second PR to change the `std::map` to `std::vector`. In my experience, `std::map` is very rarely the correct data structure to use. Often `std::unordered_map` or a sorted `std::vector` is better.

Sealer/sealer.cc
```diff
@@ -603,38 +602,48 @@ void kRun(const ConnectPairsParams& params,
...
+	int counter = 0;
+	vector<map<FastaRecord, map<FastaRecord, Gap> >::iterator> flanks_closed;
+#pragma omp parallel private(read1_it, read2_it) firstprivate(counter)
+	for (read1_it = flanks.begin(); read1_it != flanks.end(); ++read1_it) {
 		FastaRecord read1 = read1_it->first;
-		for (read2_it = flanks[read1].begin(); read2_it != flanks[read1].end(); read2_it++) {
+		bool success = false;
+		for (read2_it = flanks[read1].begin(); read2_it != flanks[read1].end(); ++read2_it, ++counter) {
+#if _OPENMP
+			if (counter % omp_get_num_threads() != omp_get_thread_num())
```
sjackman: Do you know whether these function calls are faster? Should their result be explicitly cached? I don't know the answer. I prefer this code as is if they are fast (ideally inline).

Sealer/sealer.cc
sjackman: Bump.

Sealer/sealer.cc
sjackman: Still needs addressing.

schutzekatze: There are 2 data structures here - flanks_closed and flanks. I am iterating flanks_closed that holds iterators that belong to flanks, and delete the data they point to. This should be valid, as I am not traversing flanks here, no? Before this pull request, this was done while traversing flanks and closing gaps (but in a safe way, by moving the iterator, then deleting what it previously pointed to), but I had to change it because there are now multiple threads working with flanks, so it's only safe to alter flanks after the work is done.

The reason, functionality wise, for erasing flanks, is because after kRun finishes, it proceeds to the next value of k and if they weren't erased, this new value of k could potentially close the same gaps from the previous run. At least that is my understanding, but in any case, the original code of Sealer was erasing flanks that belong to closed gaps, so I wanted to keep the functionality.

schutzekatze: So that means the original Sealer, before this pull request, was potentially not dealing with it safely either, as it deleted iterators on the go. Sounds like flagging processed flanks would be the safe way to do it, then.

Sealer/sealer.cc
schutzekatze: Acording to https://stackoverflow.com/a/22962390 
It should be slightly faster to explicitly cache it, because the functions aren't inline, so I'll see to add that.

sjackman: I'm okay with the way it is now.

benvvalk:
Thanks for your careful review, @sjackman! I will test these changes tomorrow afternoon and merge if everything looks okay.

Sealer/sealer.cc
```diff
@@ -603,38 +602,49 @@ void kRun(const ConnectPairsParams& params,
...
-			flanks.erase(read1_it++);
+#pragma omp critical (flanks_closed)
+			flanks_closed.push_back(read1_it);
 		}
-		else
-			read1_it++;
+	}
+
+	for (vector<map<FastaRecord, map<FastaRecord, Gap> >::iterator>::iterator it = flanks_closed.begin();
+			it != flanks_closed.end(); it++) {
```
sjackman: Please use preincrement `++it`

Sealer/sealer.cc
sjackman: Ah, I scanned the code too quickly! I missed that `flanks_closed` is a new data structure, and that it's a vector of iterators. I'm okay with the code as is. For a future PR to tidy up the code, I'd prefer that flanks were a vector of tuples, rather than a map of maps as it is now.

sjackman:
Please squash the 8 commits of this PR down to a single commit.
Very close now!

sjackman:
@schutzekatze This PR currently has two commits, one of which is a merge commit. Please squash and rebase this PR down to a single commit. No merge commits, please.
See https://github.com/bcgsc/abyss/pull/249/commits

schutzekatze:
Just a note - I had to revert the #omp critical (logStream) back to #omp critical (gapsclosed), because it caused a deadlock: there is a printLog call within the critical section which locks the section too.

benvvalk:
@schutzekatze Nice! I had another look through your code just now and it looks good.  I'm running my own tests on a C. elegans dataset right now, as a final check.

benvvalk:
@schutzekatze Will check my results tomorrow morning and merge if all looks well.

sjackman:
Five times faster is an excellent result! Thanks, Vladimir!

sjackman:
If you have more time in your summer hackathon (though I suspect you don't) limiting the number of vertices searched to the Sealer graph search would be excellent.

benvvalk:
@schutzekatze Merged and thank you!  (My test jobs are still running, but the log stats are identical, so good enough.)

sjackman:
Thank you very much for your contribution to ABySS, Vladimir! 🏅

schutzekatze:
Hey, that was pretty fun. To be fair, I learned a lot studying how ABySS works, good job to you for making it. Reading its papers was also pretty interesting, so, cheers!

benvvalk:
@mmokrejs Sorry to hear about the long run times.  There is an outstanding Sealer issue with long runtimes on low k values.  As a workaround, I would recommend excluding lower k-values (e.g. <= 60).

#247 has been accepted and merged into `master`.  Pending the next ABySS release, you can compile the HEAD commit of this repo to use the latest features implemented by Vladimir (i.e. Sealer gap file and Sealer multithreading).

sjackman:
I hope that limiting the number of vertices searched in the Sealer graph search, and stopping the search when it's exceeded, would improve the run time at small values of *k*.

schutzekatze:
Yeah, I have noticed that a certain number of gaps dominate the running time of Sealer. Most of the threads do their work, and then wait for a while on a thread or two, so it's definitely an issue worth looking into.

mmokrejs:
@benvvalk I am sorry for for being unclear, I forgot to say this job is running Vladimir's binary, so with all the patches applied. A periodic output of the number of gaps acted upon and to be still acted upon by each worker thread would be really helpful here. Moreove, i think the threads shoudl not get assigned the tasks at the very beginning but better could periodically fetch `.next(100)` from a joint stack. Then the CPU cores would be much more evenly loaded.

> I hope that limiting the number of vertices searched in the Sealer graph search, and stopping the search when it's exceeded, would improve the run time at small values of k.

@sjackman Would this eventually prevent long gaps to be closed at all (like in 2-8kbp range)? It seems Sealer is closing practically only gaps in the 1-700bp range although we do have two 4kbp mate-pair datasets and one 8kbp mate-pair dataset. I think Sealer or actually abyss could maybe come up with some average/median number of de Bruijn nodes needed to yield say 1kbp sequence. That could be used to define a threshold in Sealer, for example if Sealer traversed 50x more of nodes for an estimated 1kbp gap then the gaps could be given up.

@schutzekatze Would you mind at some time uploading your updated figure(s) of estimated vs. actually sealed gap sizes (with the absolute counts in legend) from your recent talk here? :-) Maybe also one showing the histogram of errors (estimated vs. real gap sizes)? Well, take your time.

I also discussed with Vladimir that currently you recommend defining k-mer values as (min, max) range while in the past you proposed (max, min) range. I expressed my opinion that in the (min, max) way we close a gap ASAP (while maybe underestimating gap size and make unwanted compression of say a tandem repeat) while in (max, min) way we are prone to fail to seal very same gap in several iterations but practically having better chance to get the number of tandem repeats correct. Also, I had the impression Sealer could keep track of the coverage at each attempted k-mer value and maybe somehow estimate whether a shift from say `k=125` to `k=90` could give a high-enough coverage so that the gap could be closed at least in theory. Maybe Sealer could skip some gaps at certain (too close) k-mer values if previous attempts using longer k-mers failed due to low coverage? Just brainstorming, not that I would be able to implement it myself.

sjackman:
> Moreover, I think the threads should not get assigned with all the tasks at the very beginning but better they could periodically fetch .next(100) from a joint stack in batches. Then the CPU cores would be much more evenly loaded.

I believe OpenMP can handle this sort of scheduling on its own. I'd need to look into it more.

sjackman:
> Would this eventually prevent long gaps to be closed at all (like in 2-8kbp range)? It seems Sealer is closing practically only gaps in the 1-700bp range although we do have two 4kbp mate-pair datasets and one 8kbp mate-pair dataset.

In `SimpleGraph` the default value for `maxCost` is 100,000 vertices. See https://github.com/bcgsc/abyss/search?q=maxCost

sjackman:
> I also discussed with Vladimir that currently you recommend defining k-mer values as (min, max) range while in the past you proposed (max, min) range. I expressed my opinion that in the (min, max) way we close a gap ASAP (while maybe underestimating gap size and make unwanted compression of say a tandem repeat) while in (max, min) way we are prone to fail to seal very same gap in several iterations but practically having better chance to get the number of tandem repeats correct.

Larger values of *k* have more missing *k*-mers due to a lack of coverage, and missing *k*-mers can cause misassemblies. So in my experience larger values of *k* are more likely to misassemble a gap, which is why we now recommend going from smaller values of *k* up to larger values of *k*.

benvvalk:
@mmokrejs @schutzekatze FYI, I believe the problem of unreasonably long Sealer (and Konnector) running times is now fixed in release [ABySS-2.1.2](https://github.com/bcgsc/abyss/releases/tag/2.1.2).

PR #257, opened 2018-10-05
Dockerfile: Add a user for OpenMPI

sjackman:
OpenMPI does not like to be run as root.


sjackman:
@benvvalk Don't merge this PR just yet. I'll test it tomorrow.

benvvalk:
@sjackman Thank you, Shaun!

sjackman:
It works!

PR #278, opened 2019-02-17
Fix failure to build on 32-bit architectures

mr-c:
Version 1.5.2-1 of abyss fails to build from source on 32-bit architectures (armhf, i386 and powerpc). Previous versions built fine on all architectures. . There were a couple of places in the code that was introduced in version 1.5.2 where size_t data type was used instead of a fixed size data type. The patch builds on amd64, i386 and armhf with and all unit tests passed.

### DIFF
--- a/Common/StringUtil.h
+++ b/Common/StringUtil.h
@@ -178,7 +178,7 @@ bool isReadNamePair(const std::string& name1, const std::string& name2)
 	return false;
 }
 
-static inline size_t SIToBytes(std::istringstream& iss)
+static inline unsigned long long SIToBytes(std::istringstream& iss)
 {
 	double size;
 	std::string units;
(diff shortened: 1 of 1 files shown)


sjackman:
Each call to `SIToBytes` is stored in a `size_t`, so `SIToBytes` should return a `size_t`.
See https://github.com/bcgsc/abyss/search?q=SIToBytes&unscoped_q=SIToBytes

Do you have a link to the original Debian bug report?

mr-c:
@sjackman http://bugs.debian.org/780409

mr-c:
(patch isn't mine, I'm just forwarding it: https://salsa.debian.org/med-team/abyss/blob/master/debian/patches/abyss-32-bit.patch )

sjackman:
I suspect that perhaps this patch isn't needed. I'm happy to revisit the issue if the original author of the patch would like to comment here with the exact compiler error they were addressing with this patch.

PR #279, opened 2019-02-17
big endian fixes from Debian

mr-c:


sjackman:
I've merged commit 5f691faf3c085375f33d6f473fdf3d9470fa18e5 `Fix FTBFS on mips and other big-endian architectures`.
I've skipped commit 350942fdc9e782144bc0fa9f07bc687486c5dfda `Fix strict alignment violation on sparc64`.
Adding an additionally `memcpy` would slow down these inner-loop functions. Architectures without hardware support for unaligned memory access are currently unsupported by ABySS.

PR #280, opened 2019-02-17
Script uses bash syntax so request bash interpreter

mr-c:


DataBase/abyss-db-txt
```diff
@@ -1,4 +1,4 @@
-#! /bin/sh
+#! /bin/bash
```
sjackman:
```suggestion
#!/bin/bash
```

PR #281, opened 2019-02-17
Fix build on hurd due to missing header file

mr-c:
https://bugs.debian.org/920240

### DIFF
--- a/Common/MemoryUtil.h
+++ b/Common/MemoryUtil.h
@@ -7,8 +7,14 @@
 #include <unistd.h> // for sbrk
 
 #if __MACH__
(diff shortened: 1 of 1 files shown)


Common/MemoryUtil.h
```diff
@@ -7,8 +7,14 @@
 #include <unistd.h> // for sbrk
 
 #if __MACH__
-# include <mach/mach.h> // for mach_task_self
-# include <mach/task.h> // for task_info
+# ifdef __APPLE__
```
sjackman:
```suggestion
# if __APPLE__
```

PR #284, opened 2019-03-14
Fixed small typo on the command for paired de Bruijn graph

sturgman:
The text around the code snippet used 64 as the span whereas the command used 96.


sjackman:
Thank you for this documentation fix, @sturgman !

sjackman:
Running ABySS 2.0 in Bloom filter mode is recommended over using ABySS's paired dBG mode (just in case). Thanks for the documentation fix all the same!

PR #287, opened 2019-06-21
Fix build on big-endian FreeBSD

pkubaj:
FreeBSD uses sys/endian.h header for byte-swapping and bswap*(x) functions.


sjackman:
Merged. Thank you for your contribution to ABySS, @pkubaj ! 🌟

PR #289, opened 2019-07-11
Set up CI with Azure Pipelines

lcoombe:


lcoombe:
The gcc6 test passed on the azure side, but was still showing as pending on the github side. I launched a re-build of the tests on azure which should hopefully have all the tests pass on this side.

sjackman:
All the tests have passed ✅! 🥳

sjackman:
We could update clang to version 8.0.0 in a future PR.

PR #330, opened 2020-04-24
RollingHash.h: report same hash as ntHashIterator

jowong4:


sjackman:
Could you please add a unit test for this bug?

PR #332, opened 2020-04-28
Unify Bloom filters in ABySS with BTL bloom filters from other projects

jowong4:
Motive: make bloom filters generated in other BTL projects compatible with abyss. 
This PR has two components: a compatible bloom filter and using the same hashing algorithm (nthash).

Previously, I attempted to accomplish the same objective by modifying the code to use BTL bloom filter, but that was unsuccessful. This time, I wrote a new class that inherits BTL bloom filter and has the same API as Konnector::BloomFitler called KonnectorBloomFilter.

Change in assembly metrics (correctness) should be expected due to change in hashes but should be negligible. There is both a gain and loss in time performance. Insertion now capitalizes on rolling hash advantage but checking doesn't. So Sealer, which doesn't check as much will see an improvement in time performance but konnector won't.  To modify the checking to capitalize on rolling hash requires more in depth code knowledge in Sealer and Konnector.

We've also decided to remove hashSeed and window from bloom.cc.


jowong4:
Assembly:
|Assembly|NG50|NGA50|# misassemblies|Genome fraction (%)|
|--|--|--|--|--|
|HG004_pre change|1,601,303|1,321,452|230|86.430|
|HG004_post change|1,600,560|1,320,826|211|86.431|
|NA12878_pre change|1,175,299|1,131,288|464|91.824|
|NA12878_post change|1,181,693|1,136,091|459|91.823|

Less than 1% difference

jowong4:
|Assembly|NG50|NGA50|# misassemblies|Genome fraction (%)|time (h:mm)|
|--|--|--|--|--|--|
|hsapiens_sealed_pre|1,592,692|1,316,708|233|86.512|9:48
|hsapiens_sealed_post|1,592,692|1,316,708|233|86.512|8:36

sjackman:
An 8% decrease in misassemblies in HG004 is significant. Thats sounds more than expected due to random variation to me. You could test that hypothesis empirically if you wanted by making multiple assemblies (say five) with different hash seeds, and estimating the mean and variance of the number of misassemblies (and for contiguity and completeness).

I haven't reviewed the PR, but I'm all for using a single compatible Bloom filter implementation across all BTL projects, ideally with a single compatible file format for the Bloom filters.

lcoombe:
> Sealer k80 benchmark:
> 
> Assembly	NG50	NGA50	# misassemblies	Genome fraction (%)	time (h:mm)
> hg004_sealed_pre	1,592,692	1,316,708	233	86.512	9:48
> hg004_sealed_post	1,592,692	1,316,708	233	86.512	8:36

Could you please add % gaps closed to this table, and the data used? (Reads, assembly)

jowong4:
Konnector
|  |time (h:mm)|
|--|--|--|--|--|--|--|--|
|pre_change|6:18|
post change|21:01|
ecoli reads
Working on a version that utilizes rolling hash

jowong4:
ABySS-bloom:
| version |k|time (h:mm)|
|--|--|--|
|pre_change|80|3:12|
|post_change|80|2:44|
|pre_change|96|3:11|
|post_change|96|2:41|
|pre_change|112|3:08|
|post_change|112|2:43|

NA12878 PE and mpet reads

sjackman:
A three-fold increase in run time for Konnector doesn't seem good. It was using rolling hash for lookup before, and no longer is, is that the problem?

jowong4:
> A fifteen-hour and three-fold increase in run time for Konnector doesn't seem good. It was using rolling hash for lookup before, and no longer is, is that the problem?

Rather, it was using citi hash to do look up which is faster than ntHash if you were to rehash the values every time which is what's currently being done. This is easier to implement at this time. I currently have a version that doesn't rehash and I'm testing that at the moment.

lcoombe:
Shh stalebot

jowong4:
I could not achieve the same performance as konnector before the changes. It's currently at around 1.5x the time which is much better than before. What's mostly likely going to happen if I don't make any breakthrough today is I will add the current code and push it. Enough time has been spent and I'm not making any headway. Will probably revisit on a later date when I have no other pending tasks.

jowong4:
Assembly Completion time
| Run |time (h:mm)|
|--|--|
|HG004 pre change|15:00|
|HG004 post change|13:59|
|NA12878 pre change|14:50|
|NA12878 post change|13:41|


Due to the optimizations I made, assembly is ~roughly an hour faster than before. Do note that the code only affects the generation of `prefix-1.fa` which means the speed increase is more substantial than ~6.6%. Currently obtaining the run time of just `prefix-1.fa`

jowong4:
Konnector results:
Dataset: Ecoli.
| Reads |count|
|--|--|
|Unique pre change |214|
|Unique post change|268|
|shared|7,436,444|

So the majority of the konnected reads are the same. Less than 0.01 % of the reads are different which I think we can chalk up to hash collisions.

Time:

| version |time (h:mm)|
|--|--|
|pre_change|14:44|
|post change|23:07|

jowong4:
Sealerr results:
Dataset: Hsapiens. hg004
| scaffolds |count|
|--|--|
|Unique pre change |61|
|Unique post change|176|
|shared|27,085|

So the majority of the konnected reads are the same. Roughly 1% of scaffolds are different which is probably due to the same reason the konnector results are slightly different.

Time:

| version |time (h:mm)|
|--|--|
|pre_change|1:53|
|post change|2:24|

||Assembly||NG50||NGA50||# misassemblies||Genome fraction (%)||
|pre_hg004_sealed_scaffold|1,598,881|1,313,474|212|86.470|
|post_hg004_sealed_scaffold|1,598,881|1,313,474|212|86.473|

jowong4:
Sealerr results:
Dataset: Hsapiens. na12878
| scaffolds |count|
|--|--|
|Unique pre change |73|
|Unique post change|148|
|shared|46,897|

So the majority of the sealed scaffolds are the same. Roughly 0.5% of scaffolds are different which is probably due to the same reason the konnector results are slightly different.



|Assembly|NG50|NGA50|# misassemblies|Genome fraction (%)|time (h:mm)|gaps found|gaps sealed|
|--|--|--|--|--|--|--|--|
|pre_na12878_sealed_scaffold|1,181,959|1,135,588|457|91.935|2:39|337,558|46,970|
|post_na12878_sealed_scaffold|1,181,959|1,135,588|457|91.935|4:51|337,558|47,045|

75 more gaps sealed

PR #335, opened 2020-05-25
abyss: detect 'ar' tool via AM_PROG_AR

trofi:
Before the change
    $ ./configure --host=x86_64-pc-linux-gnu && make
was referring to 'ar' tool.

After the change it refers to 'x86_64-pc-linux-gnu-ar'
tool if that exists. This helps user to consistently
pick 'ar' implementation.

Reported-by: Agostino Sarubbo
Bug: https://bugs.gentoo.org/725300

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ Makefile
 Makefile.in
 README.html
 aclocal.m4
+ar-lib
 autom4te.cache
 config.guess
 config.h
--- a/configure.ac
+++ b/configure.ac
@@ -22,6 +22,7 @@ AC_CHECK_TOOL(GHC, ghc)
 AM_CONDITIONAL([HAVE_GHC], ["$GHC" --version])
 AC_CHECK_PROG(PANDOC, pandoc, yes)
 AM_CONDITIONAL([HAVE_PANDOC], [test x"$PANDOC" = x"yes"])
+AM_PROG_AR
 
 # Checks for header files.
 AC_CHECK_HEADERS([dlfcn.h fcntl.h float.h limits.h \


sjackman:
```
ar cru libcommon.a libcommon_a-ContigID.o libcommon_a-Fcontrol.o libcommon_a-Histogram.o libcommon_a-Kmer.o libcommon_a-Log.o libcommon_a-Options.o libcommon_a-Sequence.o libcommon_a-SignalHandler.o libcommon_a-Timer.o libcommon_a-Uncompress.o libcommon_a-city.o 
make[2]: ar: No such file or directory
```
https://bugs.gentoo.org/attachment.cgi?id=641704&action=edit

Is this to support cross compiling?

.gitignore
```diff
@@ -2,6 +2,7 @@ Makefile
 Makefile.in
 README.html
 aclocal.m4
+ar-lib
```
sjackman, line 5: What is this file, and should it be included in the source tarball?

sjackman: Does `make dist` include `ar-lib` in the source tarball automatically? I'm guessing yes.

trofi: `ar-lib` says it's a `# Wrapper for Microsoft lib.exe`. My guess it;s a way to provide an `ar`-like interface on MSVC systems.

It should gt into tarball automatically (unless `abyss` does some unusual overrides). I'll double-check.

trofi:
> Is this change to support cross compiling? If not cross compiling, specifying --host is not necessary and should not be provided.

Either cross-compiling or picking a specific toolchain if many native toolchains are available. Say:
- both 32-bit (i686-pc-linux-gnu) and 64-bit (x86_64-pc-linux-gnu)
- or multiple contemporary libcs are installed: glibc (x86_64-pc-linux-gnu) and musl (x86_64-unknown-linux-musl)
- or multiple toolchains with small difference: x86_64-pc-linux-gnu / x86_64-hardenedstack-linux-gnu

.gitignore
trofi: Tested `make dist` today. `ar-lib` gets into tarball as expected.

sjackman:
@jowong4 I'm fine with this change. Want to merge it?

jowong4:
@sjackman  my apologies, I wanted to understand this PR before merging it, but I'm a bit preoccupied at the moment. Should have some time to do it on Friday.

jowong4:
@trofi  Thanks.

PR #337, opened 2020-05-30
Capture all tags in sam file

jowong4:
SAMtools does not set the proper pair flag if the reads are mapped and oriented correctly. We need this flag to be set for tool development. abyss-fixmate does set the proper pair flag if the reads are mapped and oriented correctly so we've decided to use abyss-fixmate but abyss-fixmate does not capture all tags.

This PR contains changes so that all tags are captured.

### DIFF
--- a/Common/SAM.h
+++ b/Common/SAM.h
@@ -340,8 +340,8 @@ struct SAMRecord : SAMAlignment {
 #if SAM_SEQ_QUAL
 		in >> o.seq >> o.qual;
 		getline(in, o.tags);
-#endif
-#if !SAM_SEQ_QUAL
+		o.tags = o.tags.substr(1);
+#else
 		in.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
 #endif
 		if (!in)
--- a/Common/SAM.h
+++ b/Common/SAM.h
@@ -339,15 +339,7 @@ struct SAMRecord : SAMAlignment {
 			>> o.cigar >> o.mrnm >> o.mpos >> o.isize;
 #if SAM_SEQ_QUAL
 		in >> o.seq >> o.qual;
-		std::string allTags = "";
-		if (in.peek() != '\n')
-			in >> allTags;
-		while (in.peek() != '\n' && !in.eof()) {
-		    std::string tag = "";
-			in >> tag;
-			allTags += "\t" + tag;
-		}
-		o.tags = allTags;
+		getline(in, o.tags);
 #endif
 		in.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
 		if (!in)
--- a/Common/SAM.h
+++ b/Common/SAM.h
@@ -339,8 +339,7 @@ struct SAMRecord : SAMAlignment {
 			>> o.cigar >> o.mrnm >> o.mpos >> o.isize;
 #if SAM_SEQ_QUAL
 		in >> o.seq >> o.qual;
-		getline(in, o.tags);
-		o.tags = o.tags.substr(1);
+		getline(in >> Skip('\t'), o.tags);
 #else
 		in.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
 #endif
--- a/Common/SAM.h
+++ b/Common/SAM.h
@@ -341,7 +341,9 @@ struct SAMRecord : SAMAlignment {
 		in >> o.seq >> o.qual;
 		getline(in, o.tags);
 #endif
+#if !SAM_SEQ_QUAL
 		in.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
+#endif
 		if (!in)
 			return in;
 		o.pos--;
--- a/Common/SAM.h
+++ b/Common/SAM.h
@@ -339,8 +339,15 @@ struct SAMRecord : SAMAlignment {
 			>> o.cigar >> o.mrnm >> o.mpos >> o.isize;
 #if SAM_SEQ_QUAL
 		in >> o.seq >> o.qual;
+		std::string allTags;
 		if (in.peek() != '\n')
-			in >> o.tags;
+			in >> allTags;
+		while (in.peek() != '\n' && !in.eof()) {
+		    std::string tag;
+			in >> tag;
+			allTags += "\t" + tag;
+		}
+		o.tags = allTags;
 #endif
 		in.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
 		if (!in)
--- a/Common/SAM.h
+++ b/Common/SAM.h
@@ -339,11 +339,11 @@ struct SAMRecord : SAMAlignment {
 			>> o.cigar >> o.mrnm >> o.mpos >> o.isize;
 #if SAM_SEQ_QUAL
 		in >> o.seq >> o.qual;
-		std::string allTags;
+		std::string allTags = "";
 		if (in.peek() != '\n')
 			in >> allTags;
 		while (in.peek() != '\n' && !in.eof()) {
-		    std::string tag;
+		    std::string tag = "";
 			in >> tag;
 			allTags += "\t" + tag;
 		}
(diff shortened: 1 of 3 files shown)


Common/SAM.h
```diff
@@ -339,8 +339,15 @@ struct SAMRecord : SAMAlignment {
...
+		std::string allTags = "";
 		if (in.peek() != '\n')
-			in >> o.tags;
+			in >> allTags;
+		while (in.peek() != '\n' && !in.eof()) {
+		    std::string tag = "";
+			in >> tag;
+			allTags += "\t" + tag;
+		}
+		o.tags = allTags;
```
sjackman, line 350: I believe that you can replace this entire chunk with
```c++
getline(in, o.tags);
```

There are some style red flags here. `!in.eof()` is big ⚠️. Usually you should instead use either `in` in a bool context (like `while (in)`) or more explicitly `!in.fail()`. The former bool context is more idiomatic.

Allocating a `std::string` on the heap in an an inner loop on line 346 is a performance / efficiency no no.
String concatenation which uses the heap in an inner loop on line 348 is a performance / efficiency no no.

jowong4: Yeah I figured, it wasn't the best implementation which is I waited for your input before I merged. Can you explain your comment about why string concatenation in the inner loop is bad?

sjackman: You want to avoid unnecessary heap allocations and frees in an inner loop when possible. The operation `allTags +="\t" + tag` allocates a new string, copies `"\t"` to it and then resizes that string and copies `tag` to it, resizes `allTags`, and copies the temporary string to it, then frees the temporary string, then frees `tag`. That's a lot of unnecessary work and access to the heap in an inner loop. Accesses to the heap in multithreaded code may involve locks and contention.

Common/SAM.h
```diff
@@ -339,15 +339,7 @@ struct SAMRecord : SAMAlignment {
...
-		std::string allTags = "";
-		if (in.peek() != '\n')
-			in >> allTags;
-		while (in.peek() != '\n' && !in.eof()) {
-		    std::string tag = "";
-			in >> tag;
-			allTags += "\t" + tag;
-		}
-		o.tags = allTags;
+		getline(in, o.tags);
```
sjackman, line 342: LGTM. Are there any places in the code that assumes that `.tags` contains only a single tag?

Common/SAM.h
```diff
@@ -339,8 +339,7 @@ struct SAMRecord : SAMAlignment {
 			>> o.cigar >> o.mrnm >> o.mpos >> o.isize;
 #if SAM_SEQ_QUAL
 		in >> o.seq >> o.qual;
-		if (in.peek() != '\n')
-			in >> o.tags;
+		getline(in, o.tags);
 #endif
 		in.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
```
sjackman, line 344: The tags and EOL character are extracted by `getline(in, o.tags);`, and then this line `in.ignore` will ignore the entire *next* line of the SAM input. The tests passed, which is frightening. Can you please add a unit test with multiple SAM lines and that fails with this code? After verifying that this unit test fails, you can fix this issue.

jowong4: Hmm, the file going in and the file going out has the same number of lines.

sjackman: That would be good news, but I can't reconcile that observation with the code above. `getline` consumes the end-of-line character: `the delimiter character is extracted from input`. https://en.cppreference.com/w/cpp/string/basic_string/getline
The following `in.ignore` would then skip all characters until the following end-of-line character of the next line.

sjackman: Did you test with `SAM_SEQ_QUAL` defined?

jowong4: The sam file has tags in the input and tags in the output so I'm fairly certain `SAM_SEQ_QUAL` is defined.

sjackman: I bet that the caller that you're testing has slurped one single line into an `istringstream`, and you're parsing that single line in your test. The caller is not guaranteed to be using an `istringstream` though. If it used `std::cin` you'd see every other line skipped. You could write a small unit test to test parsing an `istringstream` that contains a few lines of SAM.

Common/SAM.h
jowong4: As far as I can tell, the tags variable is only used to save the tags output.

Common/SAM.h
jowong4: I see. In any case, can I not just remove line 344 since `getline()` will consume what is remaining until the next new line character?

Common/SAM.h
sjackman: Look for other locations that extract `BX:Z:` from the SAM tags, and ensure they can handle multiple SAM tags.

Common/SAM.h
sjackman: When `SAM_SEQ_QUAL` is defined, yes. When it's not defined, you still need the `.ignore()`.

Common/SAM.h
```diff
@@ -339,10 +339,11 @@ struct SAMRecord : SAMAlignment {
 			>> o.cigar >> o.mrnm >> o.mpos >> o.isize;
 #if SAM_SEQ_QUAL
 		in >> o.seq >> o.qual;
-		if (in.peek() != '\n')
-			in >> o.tags;
+		getline(in, o.tags);
 #endif
+#if !SAM_SEQ_QUAL
```
sjackman, line 344:
```suggestion
#else
```

sjackman:
I recommend adding a unit test with two SAM records separated by a newline character in a istringstream. Ideally test with and without `SAM_SEQ_QUAL` defined.

jowong4:
@sjackman I can add the unit test. but how would I make a test without `SAM_SEQ_QUAL` defined? The variable is set in compilation time. Would I sed the file to change the variable and compile?

sjackman:
https://github.com/bcgsc/abyss/blob/5ae4e48188799bd6520b9248121e7afd985bb6f2/Unittest/Makefile.am#L39-L41

https://github.com/bcgsc/abyss/blob/5ae4e48188799bd6520b9248121e7afd985bb6f2/Map/Makefile.am#L31-L38

Add a new test that looks something like… (totally untested):
```c++
check_PROGRAMS += common_sam_ssq
common_sam_ssq_SOURCES = Common/SAM.cc
common_sam_ssq_CPPFLAGS = -DSAM_SEQ_QUAL
common_sam_ssq_LDADD = $(top_builddir)/Common/libcommon.a $(LDADD)
```

Common/SAM.h
```diff
@@ -339,10 +339,11 @@ struct SAMRecord : SAMAlignment {
 			>> o.cigar >> o.mrnm >> o.mpos >> o.isize;
 #if SAM_SEQ_QUAL
 		in >> o.seq >> o.qual;
-		if (in.peek() != '\n')
-			in >> o.tags;
-#endif
+		getline(in, o.tags);
+		o.tags = o.tags.substr(1);
```
sjackman, line 343:
```suggestion
		getline(in >> Skip('\t'), o.tags);
```
`substr` makes a copy of the string which is requires a heap allocation and is unnecessarily inefficient. `o.tags.erase(0, 1);` would be better, and the above suggestion is better still.

sjackman (changes requested):

Common/SAM.h
jowong4: ``
```suggestion
                in.ignore(1);
		getline(in, o.tags);
```
What about this change? Or is the skip version still preferred?

sjackman (approved):

sjackman:
It’s not a bug exactly in that it was a design decision that we grabbed only the first SAM tag, which contained the barcode `BX:Z:` tag that we needed, and ignored the rest because we didn’t need them. That probably should have been explained in a comment. If you don’t need the rest of the tags, then what’s the motivation to extract them? It’s a hack to assume that the `BX:Z:` comes first, to be sure.

PR #354, opened 2020-12-02
Change Linuxbrew to Homebrew

sjackman:
The Homebrew and Linuxbrew projects have merged.


lcoombe:
Thanks for this Shaun!

sjackman:
Just a heads up that https://linuxbrew.sh will going away soon. I recommend linking instead to either https://brew.sh or https://docs.brew.sh/Homebrew-on-Linux.

PR #373, opened 2021-07-20
small genomes too

afshinfard:
As in our latest experiments abyss is doing just as nice on smaller genomes, we decided to include in the description that it does work on smaller genomes too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -586,8 +586,8 @@ Nielsen, Cydney B., Shaun D. Jackman, Inanc Birol, and Steven JM Jones (2009).
 Support
 =======
 
-[Ask a question](https://www.biostars.org/p/new/post/?tag_val=abyss,assembly)
-on [Biostars](https://www.biostars.org/t/abyss/).
+[Ask a question](https://www.biostars.org/new/post/?tag_val=abyss,assembly)
+on [Biostars](https://www.biostars.org/tag/abyss/).
 
 [Create a new issue](https://github.com/bcgsc/abyss/issues) on GitHub.
 
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ ABySS
 =====
 
 ABySS is a *de novo* sequence assembler intended for short paired-end
-reads and large genomes, but it also works well for small (e.g. bacterial) genomes.
+reads. Initially designed for large genomes, ABySS also performs a decent job assembling small (e.g. bacterial) genomes.
 
 Please [cite our papers](#citation).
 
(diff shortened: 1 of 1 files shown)


afshinfard:
Also, the `biostar` links does not work anymore. We can fix them too. Do I change to the Biostar homepage or that was some sort of a feature still available but in a different way?

lcoombe:
> Also, the `biostar` links does not work anymore. We can fix them too. Do I change to the Biostar homepage or that was some sort of a feature still available but in a different way?

I think it just needs to be updated to https://www.biostars.org/tag/abyss/ ?

afshinfard:
> I think it just needs to be updated to https://www.biostars.org/tag/abyss/ ?

Right, and I also removed `/p/` in the other link and it is working now.

afshinfard:
@lcoombe I do not have access to add reviewers, labels, etc, and also to merge. If you think this PR is ready let's make it to the master. Please note I edited the intro sentence one more time to make sure the message is clear that abyss is really good also for smaller genomes; sounds good?

README.md
```diff
@@ -7,7 +7,7 @@ ABySS
 =====
 
 ABySS is a *de novo* sequence assembler intended for short paired-end
-reads and large genomes.
+reads. Initially designed for large genomes, ABySS also performs a decent job assembling small (e.g. bacterial) genomes.
```
lcoombe, line 10:
```suggestion
reads. Initially designed for large genomes, ABySS also performs well on small (e.g. bacterial) genomes.
```
Or some thing like that? I wasn't sure about the phrase "decent job"

afshinfard: My only concern is that people may read this and think "well, they mean it does work on smaller genomes but probably they're not the best". But we know we're doing as good as other alternatives... so how do you think we can make this impression?

afshinfard: @lcoombe, I've seen that on some well-known repos people add a comparison too.. like a small table showing their results side-by-side of another tool well-used by the community... that'd show everything but I am not sure if we like that sort of thing in a repo readme

sjackman: ABySS is a *de novo* sequence assembler intended for short paired-end reads and genomes of all sizes.

lcoombe: Thanks Shaun - I like this better!
I think updating that text is fine for now, if desired we could add more in a future PR.

afshinfard: Thanks @sjackman, I like your idea! It's simple and right! we don't need to give a background of what abyss was designed for in here!

PR #464, opened 2023-04-28
Make btllib an external dependency

lcoombe:
* `btllib` is currently packaged within ABySS, and required for Rresolver
* This PR separates out the `btllib` code so that it is a dependency vs. packaged within ABySS
  * Better code separation, and easier to adapt to newer btllib release versions
* Ported the tests to Azure pipelines only, removed circleCI
  * Consistent with most of our other repositories
  * Some older tests were removed due to btllib version requirements


jwcodee (approved):
Thabks

Issue #466, opened 2023-05-06
Build failure on macOS 12 and 13


p-linnane:
Hello 👋 . I'm a maintainer for the [Homebrew](https://brew.sh) project. While packaging v2.3.6 of abyss we are running into build failures. The error is as follows:

```
  ld: in ../DataBase/libdb.a(libcommon.a), archive member 'libcommon.a' with length 49632 is not mach-o or llvm bitcode file '../DataBase/libdb.a' for architecture x86_64
  collect2: error: ld returned 1 exit status
  make[1]: *** [ABYSS] Error 1
  make: *** [install-recursive] Error 1
```

The error is also present on arm64 architecture. The relevant GitHub Actions run can be found [here](https://github.com/Homebrew/homebrew-core/actions/runs/4903089549/jobs/8755561629?pr=130187).

I'd also like to add that @parham-k was a huge help in getting btllib working for the formula!

Relates to https://github.com/Homebrew/homebrew-core/pull/130187

lcoombe:
Hi @p-linnane,

Thanks for reaching out! I have seen that issue before in our CI, and only when using the VM mac-latest (12). The same commands run fine with macOS-11.

I took a look and it looks like similar errors have been reported elsewhere in homebrew?
https://github.com/Homebrew/homebrew-core/pull/123275 
https://github.com/Homebrew/homebrew-core/pull/113261

Looks like there was an attempted patch - https://github.com/Homebrew/homebrew-core/blob/c4d5a7d38d8e1761819a72249c5232dda8f76647/Formula/libpinyin.rb#L4

Do you think it's worth trying the same fix for the abyss brew formula?

Bo98:
This is a build script bug.

https://www.gnu.org/software/automake/manual/html_node/Program-and-Library-Variables.html

`LIBADD` is only permitted to contain object files but in various places in this project it is passing `.a` archives. This results in `.a` archives inside other `.a` archives, which is unlikely to be what you want. GNU linkers and older Apple linkers just happen to silently ignore those scenarios.

p-linnane:
The issue is with the linker in Xcode 14. macOS 11 cannot run higher than Xcode 13.2.

The Xcode 14 linker errors out when finding an `.a` file within another `.a` file.

fxcoudert:
The problem is in abyss: `libdb.a` should contain only object files (`*.o`) and not another static archive (`libcommon.a`).

The issue is here:
[...]

PR #468, opened 2023-05-16
Fix build failure on macOS 12 and 13

parham-k:
Addresses #466

- Replaced linking to other `.a` files in libraries with lists of `.o` files.
- Re-added compiling with GCC-12 on macOS-latest to CI.


lcoombe (approved):
Thanks so much Parham!

PR #471, opened 2023-08-31
Update README.md

lcoombe:
* We've noticed locally some issues can come up if conda-forge isn't part of the channel list 
  * Just adding it in the conda install command in case
 * Had to make some CI tweaks for the test to pass


jwcodee (approved):
Thanks.

PR #475, opened 2024-03-08
configure: simplify check for GHC

eli-schwartz:
This tool lookup and included AM_CONDITIIONAL was originally added in commit 73b5e40098b56b982d45e491e149d22ae7f0099a, using the same logic introduced in this commit. It was originally changed in commit 95558c6472ea52573c3487229de5ef7e1ec979d6 to avoid checking whether GHC was found, and instead check whether ghc-pkg claimed mmap was found.

A long time later, the use of ghc-pkg was removed in commit e7a3741f455a9d6e3a4a6bddcb411097fe9e7036, and the check was restored to "check if GHC exists", which however did not restore the *mechanism* for checking.

Instead, it attempted to run GHC, whether it existed or not. Checking if it was found should be enough. But checking that it runs --version is not particularly useful:
- a command not found / permission denied error is produced if GHC was not found, due to trying to execute "" as a command:

  ./configure: 7028: : Permission denied

- if it was found, the version output would be interspersed into ./configure output

Similarly a couple lines below, pandoc is checked for and only its presence is cared about.


lcoombe:
Thanks for your contribution, @eli-schwartz - the CI failure isn't due to your PR, I'm also seeing failures with the current master branch. I'm troubleshooting that now - will let you know when it is solved.

eli-schwartz:
The brew install stuff looks wonderfully fun!

We had a similar issue and solved it with this: https://github.com/mesonbuild/meson/blob/master/.github/workflows/macos.yml#L87-L92

IMO that will be more reliable over time than regularly updating specific filenames.

lcoombe:
Thanks so much for pointing me to that!
Totally agree - doing the removal of specific filenames can be like a game of whac-a-mole..

eli-schwartz:
Sure, done.

lcoombe (approved):
Thanks @eli-schwartz!

Issue #474, opened 2024-03-08
Build fails with LTO


eli-schwartz:
I tried to build abyss-2.3.4 with the following *FLAGS to optimize the build: `-flto=4 -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing`

Note the -Werror=*  flags are used to help detect cases where the compiler tries to optimize by assuming UB cannot exist in the source code -- if it does exist, ordinarily the code would be miscompiled, and this says to make the miscompilation a fatal error.

I got this build error:

```
x86_64-pc-linux-gnu-gcc  -march=native -fstack-protector-all -O2 -pipe -fdiagnostics-color=always -frecord-gcc-switches -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3 -fstack-clash-protection -flto=4 -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing  -Wformat -Werror=format-security -Werror=implicit-function-declaration -Werror=implicit-int -Werror=int-conversion -Werror=incompatible-pointer-types  -Wl,-O1 -Wl,--as-needed -flto=4 -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing -Wl,--defsym=__gentoo_check_ldflags__=0 -o dialign museq.o libdialign.a -ldl -lm 
museq.c:58:13: error: type of ‘simple_aligner’ does not match original declaration [-Werror=lto-type-mismatch]
   58 | extern int  simple_aligner(struct seq_col *scol, struct diag_col *dcol,
      |             ^
assemble.c:398:6: note: return value type mismatch
  398 | char simple_aligner(struct seq_col *scol, struct diag_col *dcol,
      |      ^
assemble.c:398:6: note: type ‘char’ should match type ‘int’
assemble.c:398:6: note: ‘simple_aligner’ was previously declared here
lto1: some warnings being treated as errors
lto-wrapper: fatal error: x86_64-pc-linux-gnu-gcc returned 1 exit status
compilation terminated.
/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../x86_64-pc-linux-gnu/bin/ld: error: lto-wrapper failed
collect2: error: ld returned 1 exit status
```

Originally reported downstream: https://bugs.gentoo.org/862252
Full build log: [build.log](https://github.com/bcgsc/abyss/files/14531992/build.log)

lcoombe:
[...]

PR #479, opened 2024-07-02
Type changes required for LTO

jwcodee:
Per #474, current ABySS will not compile with LTO flags, the current changes fixes the flags as well as warnings flagged by newer version of compilers.


lcoombe (approved):
Thanks for seeing this work through, Johnathan!

CI is now fixed, and you can merge at your convenience. I can do the release once you are happy to merge.

PR #484, opened 2024-10-03
CI fixes

lcoombe:
* Recently, macOS-latest VM was updated to use macOS-14
  * This VM no longer has conda installed by default
* Workaround is to install miniforge as part of the CI
  * This already has mamba, which is a side benefit
* Added dedicated CI for clang 15, 16 ; removed old clang-12

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,116 +1,116 @@
 jobs:  
-- job: mac_latest_gcc
-  pool:
-    vmImage: macOS-latest
-  steps:
-  - script: |
-      find /usr/local/bin -lname '*/Library/Frameworks/Python.framework/*' -delete  # Trying solution from https://github.com/mesonbuild/meson/blob/master/.github/workflows/macos.yml#L87-L92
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -33,13 +33,20 @@ jobs:
   pool:
     vmImage: 'macOS-latest'
   steps:
-  - script: echo "##vso[task.prependpath]$CONDA/bin"
-    displayName: Add conda to PATH
+  - script: |
+      mkdir -p ~/miniforge3
+      curl -L https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-MacOSX-x86_64.sh  -o ~/miniforge3/miniforge.sh
+      bash ~/miniforge3/miniforge.sh -b -u -p ~/miniforge3
+      rm -rf  ~/miniforge3/miniforge.sh
```
jwcodee, line 40: is this necessary?

jwcodee (approved):

azure-pipelines.yml
jwcodee: I see. Just wondering.

Issue #486, opened 2024-10-15
DistanceEst: invalid option: `-n15-20'


NatJWalker-Hale:
# Please report

- [x] version of ABySS with `abyss-pe version`

`2.3.9`

- [x] distribution of Linux with `lsb_release -d`

`Rocky Linux release 8.10 (Green Obsidian)`

# Assembly error

- [x] complete `abyss-pe` command line

Iterating over different k and kc

```
abyss-pe j=32 k=$k kc=$kc B=50G v=-v name=assembly lib='pea' mp='mp1000 mp1500 mp2000 mp5000 mp10000 mp20000 mp50000 mp100000 mp200000' \
 pea='read1.fq.gz read2.fq.gz' \
 mp1000='mp1000_1.fq.gz mp1000_2.fq.gz' \
 etc.
```

- [x] last 20 lines of the output of `abyss-pe`

```
Read 18000000 alignments. Hash load: 4 / 53 = 0.0754717 using 680 kB.
Read 19000000 alignments. Hash load: 2 / 53 = 0.0377358 using 680 kB.
Mapped 18737560 of 19246262 reads (97.4%)
Mapped 17467721 of 19246262 reads uniquely (90.8%)
Read 19246262 alignments
Mateless         0
Unaligned    29413  0.306%
Singleton   449876  4.67%
FR            2243  0.0233%
RF         2482693  25.8%
FF            1386  0.0144%
Different  6657520  69.2%
Total      9623131
RF Stats mean: 702.9 median: 700 sd: 70.29 n: 2403099 min: 333 max: 1071 ignored: 81837
                                   ▁█▂_
make: *** [/nobackup/dbl0hpc/apps/miniconda3/envs/abyss/bin/abyss-pe.Makefile:731: mp1000-6.dist.dot] Error 1
n	n:500	L50	min	N75	N50	N25	E-size	max	sum	name
7267482	453700	118936	500	843	1368	2268	1789	16934	537.6e6	assembly-unitigs.fa
6992864	360489	72770	500	996	2315	3908	2909	27273	562.8e6	assembly-contigs.fa
error: `assembly-scaffolds.fa': No such file or directory
```
However the actual error is at the distance estimation step:

```
	|DistanceEst -v  --dot --median -j32 -k53  -l40 -s100 -n15-20  -o mp1000-6.dist.dot mp1000-6.hist
Reading from standard input...
Reading `assembly-6.fa'...
DistanceEst: invalid option: `-n15-20'
```
[...]

PR #487, opened 2024-10-18
Fix abyss-pe to pass N parameter as expected

lcoombe:
* `N` was not passed to DistanceEst as expected
  * When only PE reads specified, `-n` was propagated from the contig stage
  * When MP reads were specified, if a range was specified, this was passed to `-n` causing a failure (reported in #486)
* Solution here: Ensure that the scaffold parameters are properly passed to DistanceEst for both read inputs, using the lower number if a range is provided


jwcodee (approved):
Thanks.

PR #498, opened 2026-04-30
fix build failure with Gcc 16

emollier:
Starting with Gcc 16, unused but set variables have become error. This results in the following build failure in abyss.

	Consensus.cpp:379:18: error: variable ‘numIgnored’ set but not used [-Werror=unused-but-set-variable=]
	  379 |         unsigned numIgnored = 0;

This change drops occurrences of the numIgnored variable, since it is unused.

Bug-Debian: https://bugs.debian.org/1132058


lcoombe (approved):
Looks good to me - thank you @emollier!

emollier:
> Looks good to me - thank you @emollier!

You're welcome!

Have a nice day,  :)
Étienne.

